PR #10, opened 2021-01-05
add support for retro devices

codepoet80:
Would you consider adding a new quality type, that would allow this service to request "safe" content for retro devices? I've been testing this for the past 3 weeks on a range of devices made between 2008 and 2011 and it works flawlessly with the quality settings I've included as "retro" in this pull request.
I didn't modify the UI since a) I suck at Angular and b) retro devices need their own UI anyway.

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -36,8 +36,9 @@ class Download:
         self.download_dir = download_dir
         if quality == 'best':
             self.format = None
-        elif quality == 'retro':
-            self.format = f'bestvideo[ext=mp4]+bestaudio[ext=aac]/best[ext=mp4]/best'
+        elif 'custom' in quality:
+            quality = quality.replace('custom:', '')
+            self.format = f'{quality}'
         elif quality in ('1080p', '720p', '480p'):
             res = quality[:-1]
             self.format = f'bestvideo[height<={res}]+bestaudio/best[height<={res}]'
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -36,6 +36,8 @@ class Download:
         self.download_dir = download_dir
         if quality == 'best':
             self.format = None
+        elif quality == 'retro':
+            self.format = f'bestvideo[ext=mp4]+bestaudio[ext=aac]/best[ext=mp4]/best'
         elif quality in ('1080p', '720p', '480p'):
             res = quality[:-1]
             self.format = f'bestvideo[height<={res}]+bestaudio/best[height<={res}]'


alexta69:
Well, the whole point of this software is the UI, why would you need it if you're not using the UI? Without the UI you can just use youtube-dl directly...

codepoet80:
Hey thanks for responding. Your software is way more useful than you might be imagining.

Retro devices, and many mobile devices, can't easily run youtube-dl directly -- your software allows enthusiasts to use YouTube without needing to have a native youtube-dl built for their environment. Many of those devices can't render your web front-end either, but can make crude web requests.

As an example, I've built a simple service wrapper in PHP that serves webOS clients, allowing them to interact with MeTube, without your UI: https://github.com/codepoet80/webos-metube

All that said, none of that is in this pull request. I'm just proposing that you expose quality settings beyond what you've currently got. I'd wager there are lots of use cases could benefit. If you don't like my approach, perhaps there's another we could make the quality options extensible? I'd rather not maintain a fork, if I can contribute to your main somehow!

alexta69:
I'm very glad you find it useful! And don't mind at all extending it to suit your use case. Perhaps let's try to think how to do it generically though, so you don't need to make changes in metube for every quality. Something like, adding `elif quality.startswith("custom:")` as the last of the bunch there, and in that condition, extract the string after `custom:` and use it as the format (i.e. `self.format = quality[7:]`). And then you can send whatever string you need for the quality.

codepoet80:
I love it!
So rather than call:
`http://myserver/add?url=videotoget&quality=retro`
I would call
`http://myserver/add?url=videotoget&quality=custom:bestvideo[ext=mp4]+bestaudio[ext=aac]/best[ext=mp4]/best`
Did I get that right?

alexta69:
Exactly! you just need to URL-escape it so that it gets parsed correctly by the server, but that's the idea, yes.

codepoet80:
Perfect! Shall I submit a new pull request, or wait for an update to the main?

alexta69:
You can just push a new commit to this PR with this logic, and I'll merge it.

codepoet80 (commented):
How does this look?

alexta69:
I did the same in a form I'm more comfortable with (code-style-wise), take a look at this commit:
https://github.com/alexta69/metube/commit/62602b19c998b7869dfeb363de27df9b7a863781
I pushed it to master. It does the same, basically, so it should fit your use case.
Sorry for not using the PR after all and thanks for the contribution!
Let me know if you're OK with it, and feel free to close the PR in that case.

codepoet80:
Resolved with [62602b1](https://github.com/alexta69/metube/commit/62602b19c998b7869dfeb363de27df9b7a863781)

PR #13, opened 2021-01-12
Multi platform build

### LINKED ISSUE #12: Docker Image for ARM
Please add a Docker Image for ARM7-Platform.

fwidder:
I have added a script for building multi-platform images.
This could also be embedded in a CI platform.

It would be cool if you could run the build with "-p true" so I don't have to upload the image to Docker Hub myself ...
If you want, I'd be happy to help you create a CI pipeline for this project.  Just get in touch with me.

Many greetings
Florian

Fixes #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,12 @@ A Docker image can be built locally (it will build the UI too):
 docker build -t metube .
 ```
 
+A Multi-Platform Docker image can be built locally:
+
+```bash
+build-multi-platform.sh
+```
+
 ## Development notes
 
 * The above works on Windows as well as Linux.
--- a/README.md
+++ b/README.md
@@ -73,7 +73,17 @@ docker build -t metube .
 A Multi-Platform Docker image can be built locally:
 
 ```bash
+# Build with tag "metube"
 build-multi-platform.sh
+
+# Build with tag "other_tag"
+build-multi-platform.sh -t "other_tag"
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -5,8 +5,11 @@ COPY ui ./
 RUN npm install && \
     node_modules/.bin/ng build --prod
 
+ARG BASE_IMAGE_PREFIX
+FROM ${BASE_IMAGE_PREFIX}python:3.8-alpine
 
-FROM python:3.8-alpine
+ARG ARCH
+COPY qemu-${ARCH}-static* /usr/bin
```
alexta69, line 12: This will break the simple use case of checking out the code and running "docker build", for local development. I wonder if there's any solution possible to still support this use case.

hooks/build
```diff
@@ -0,0 +1,57 @@
...
+
+    jq --arg architecture "$manifest_arch" \
+       --compact-output '.architecture=$architecture' < "$filename" | sponge "$filename"
+  done
+
+  tar cv . | docker load
+  rm -rf $tempdir/*
+done
+
+trap "exit 1"          HUP INT PIPE QUIT TERM
```
alexta69, line 56: Don't the trap commands need to be at the top of the file, to be executed first? If you're already at the end of the execution, it seems that the traps are no longer relevant anymore, no?

alexta69:
Done using GitHub actions.

Issue #15, opened 2021-01-24
form data for /add handler


Rpsl:
Hello, thx for great UI, i use that on my NAS and it's best solution of all the ones I've tried.

After I started using it, I wanted to make chrome extension with context menu actions. Something like: right click -> send to metube -> %magick% -> video is downloaded :)

My NAS don't have SSL i can't send XHR post requests with JSON data. I can send `post` data via emulation form submit but that need make changes in `/add/` handler for support `x-www-form-urlencoded` data.

https://github.com/alexta69/metube/blob/master/app/main.py#L63

I think i can make PR with new handler or changes in `/add` handler. What do you think?

alexta69:
Hi Victor,
Thinking about it, we don't currently have CSRF protection here, and what you're suggesting is exploiting this fact to be able to make cross-site requests :) As long as we don't have anti-CSRF measures anyway, I guess I don't mind.
In the future if we add CSRF protection, we can have a flag (via an environment variable) that allows disabling it, for use cases like yours, where you need it open. So if you do this, and find that it suddenly broke at some point in the future, just remember to check if some update didn't add CSRF protection :)
So yeah, you're welcome to submit a PR. Basically what you'll want to do is inside the add() handler, check the content type, and parse the request accordingly. Should be pretty simple.
Thanks!
Alex

alexta69:
Fixed by #17

PR #17, opened 2021-01-26
Add support of CORS rules

Rpsl:
In #15 i wrote about making changes for support submitting POST data via browser extension. After little research I understood that my problem can be solved by supporting CORS rules. 

In this PR i added middleware for supporting `Origin` header in request. Also i added one more handler for `OPTIONS` pre-request which make by browser.

I tested that solution in my environment and also i create extension for Google Chrome https://github.com/Rpsl/metube-browser-extension

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -97,8 +97,23 @@ if config.URL_PREFIX != '/':
 
 routes.static(config.URL_PREFIX + 'favicon/', 'favicon')
 routes.static(config.URL_PREFIX, 'ui/dist/metube')
-
 app.add_routes(routes)
+
+
+# https://github.com/aio-libs/aiohttp/pull/4615 waiting for release
+# @routes.options(config.URL_PREFIX + 'add')
+async def add_cors(request):
+    return web.Response(text=serializer.encode({"status": "ok"}))
+
+app.router.add_route('OPTIONS', config.URL_PREFIX + 'add', add_cors)
+
+
+async def on_prepare(request, response):
+    if 'Origin' in request.headers:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -97,10 +97,25 @@ def index_redirect_dir(request):
...
 
+# https://github.com/aio-libs/aiohttp/pull/4615 waiting for release
+# @routes.options(config.URL_PREFIX + 'add')
+async def add_cors(request):
+    return web.Response(text=serializer.encode({"status": "ok"}))
+
+app.router.add_route('OPTIONS', config.URL_PREFIX + 'add', add_cors)
+
+
+async def on_prepare(request, response):
```
alexta69, line 111: Aren't the Access-Control headers only relevant for the OPTIONS request? In which case it might be better to stick them in the add_cors() method above? I'm not sure about CORS mechanics in this regard, whether it needs to be in every response or only in the response to OPTIONS.

alexta69:
Thanks for the contribution! I wrote a comment, let me know what you think.
Also I loved the Chrome extension! Very useful! Is there any chance you're going to publish it in the Chrome webstore?

app/main.py
Rpsl: No. Browser making two request:
- OPTIONS (without submiting data) to see if it's possible 
- POST (with submiting data)

Both responses should say that the request is allowed

Rpsl:
> Is there any chance you're going to publish it in the Chrome webstore?
Yep. At now i make some changes and try some scenarios and i don't want publish new version by 100500 times at day. After a while i publish it in store.

PR #31, opened 2021-05-18
environment variable to customize output file names

xiongzi:
Additional environment variable OUTPUT_TEMPLATE to set an own value for youtube-dl's -o paramter.


alexta69:
Thanks for the contribution @xiongzi !

PR #36, opened 2021-07-24
Add AUDIO_DOWNLOAD_DIR option

### LINKED ISSUE #35: Add the option of downloading to another directory for audio files
Add AUDIO_DOWNLOAD_DIR option for audio files

omeryagmurlu:
fixes #35

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM node as builder
 
 WORKDIR /metube
 COPY ui ./
-RUN npm install && \
+RUN npm ci && \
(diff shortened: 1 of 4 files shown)


omeryagmurlu:
I've also changed the `npm install` in the Dockerfile to `npm ci`, install was causing the build to fail cause of some updated dependency

alexta69:
> I've also changed the `npm install` in the Dockerfile to `npm ci`, install was causing the build to fail cause of some updated dependency

Thanks for the contribution, and for `npm ci` as well, I wasn't aware of this distinction!

Issue #25, opened 2021-03-31
If a youtube video fails to download there's no retry button


TheKnarf:

alexta69:
Another good idea! Let's do that :)

alexta69:
Fixed by #38

PR #38, opened 2021-07-29
Added retry button for failed download

Rpsl:
I added a button to retry failed download 

//cc https://github.com/alexta69/metube/issues/25

https://user-images.githubusercontent.com/265634/127456671-b4e32e51-965a-48d2-b6b1-ae41de76ad70.mov

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -25,8 +25,9 @@ class DownloadQueueNotifier:
         raise NotImplementedError
 
 class DownloadInfo:
-    def __init__(self, id, title, url):
+    def __init__(self, id, title, url, quality):
         self.id, self.title, self.url = id, title, url
+        self.quality = quality
         self.status = self.msg = self.percent = self.speed = self.eta = None
 
 class Download:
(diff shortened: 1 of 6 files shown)


alexta69:
Thanks for the contribution!
Two questions --
1. Why did you decide to use this icon from the "solid" library instead of the "regular" library used for the other icons?
2. Perhaps it makes sense to remove it from the "failed" list when it's clicked, in other words, have this click move the entry from the bottom list to the top list, and let it move down naturally back into the bottom list once the download succeeds or fails?

Rpsl:
> Why did you decide to use this icon from the "solid" library instead of the "regular" library used for the other icons?
Yeah, I've been thinking about it, too. I couldn't find the right icon in "[regular](https://fontawesome.com/v5.15/icons?d=gallery&p=2&s=regular&m=free)" icon set 🤷 

> have this click move the entry from the bottom list to the top list
Sounds good. I just didn't notice it. I fix that asap

alexta69:
The icon is only available in the "Pro" version of FontAwesome for the Regular weight.. what a bummer :(

Rpsl:
> have this click move the entry from the bottom list to the top list

[complete](https://github.com/alexta69/metube/pull/38/commits/26ae9427868a11acba31dedeab282c5f67d4c0dc)

PR #39, opened 2021-07-29
Added local dev environment via separated docker containers

Rpsl:
While i worked on another PR i need to run application in development mode but i didn't want install requirements in OS.

I create two separate docker files for dev containers, wrote docker compose file and pack commands into makefile.

### DIFF
--- a/Dockerfile-dev-backend
+++ b/Dockerfile-dev-backend
@@ -1,0 +1,18 @@
+FROM python:3.8-alpine
+
+COPY . /app
+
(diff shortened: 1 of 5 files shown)


Tuhinakhand (changes requested):
Y

Tuhinakhand (changes requested):
>

alexta69:
Closing as outdated — the project has since moved to uv/pnpm and a different build setup, so this no longer applies.

PR #55, opened 2021-09-18
feat: configurable CORS accept for socketio

tvanriel:


alexta69:
can you describe the use case for this? what problem it solves?

tvanriel:
I couldn't connect to the websocket server because it rejected connections with a 400 Bad Request with a body containing `<my-url> is not an accepted origin`

alexta69:
is your server configured in any special way? I'm asking because it works for everyone else out of the box, so I'm curious about the specific case where it fails.
thanks for the PR! I'll merge it, I just want to understand the use case first.

tvanriel:
It isn't configured in a special way.  I use nginx for SSL termination and just run under a simple subdomain.

alexta69:
I understand. it's still strange that you're seeing a problem with it not seen by other people. but we already allow CORS on the HTTP calls, so I see no reason to disallow it on the websocket.
can you change the PR so that no environment variable is used, and it's always '*'? if we allow controlling CORS from an environment variable, we should control it for the websocket and the HTTP calls together, but there doesn't seem to be a need for it right now, so we can just allow it for both.

tvanriel:
Sure!

alexta69:
one more request -- amend the commit message since it's not configurable anymore :)

PR #73, opened 2021-10-25
Added simple MP3 support

asuyou:
If quality is set to `audio only` and format is set to `MP3`, it will convert to mp3 and embed a thumbnail

I Need to:
- Add indication of conversion
- Fix it breaking if quality is set to `best` but format is `mp3`

I'm still a junior dev so please let me know if there's anything else, or if I should format things differently!


alexta69:
> Largest change is that `best` now downloads best possible quality instead of best in a specific file type which can lead to some large files...

Not sure what you mean.
Right now when "Any" is selected in Format, it won't let you set the resolution, this needs to be addressed, can you enable resolution selection when Any is selected?
Also, the Quality and Format dropdowns have switched places, any particular reason for that? I think it might confuse existing users of the program, let's not do that unless there's a necessity.

app/dl_formats.py
```diff
@@ -0,0 +1,85 @@
...
+      Exception: unknown quality, unknown format
+
+    Returns:
+      dl_format: Formatted download string
+    """
+    audio_fmt = ""
+    video_fmt = ""
+    final_fmt = ""
+
+    if format.startswith("custom: "):
```
alexta69, line 19: There's an extra space after `custom:`, needs to be removed

app/dl_formats.py
```diff
@@ -0,0 +1,85 @@
...
+        final_fmt = format[7:]
+    elif format == "any":
+        final_fmt = "bv*+ba/b"
```
alexta69, line 22: Why the change from the previous format selections? (I mean why `bv*+ba/b`?)

asuyou:
Thank you for the feedback!

> Also, the Quality and Format dropdowns have switched places, any particular reason for that? I think it might confuse existing users of the program, let's not do that unless there's a necessity.

Ahh sure, I'll switch them back around. I originally did it to make it more clear which option should be selected first. It  should make sense anyways so I'll move them back

> Right now when "Any" is selected in Format, it won't let you set the resolution, this needs to be addressed, can you enable resolution selection when Any is selected?

I'll work on that

> > Largest change is that best now downloads best possible quality instead of best in a specific file type which can lead to some large files...

> Not sure what you mean.

(I might not of understood yt_dlp properly) What I thought was happening originally was that it was downloading the best file with the mp4 extension, whilst this one downloads the best possible file type (e.g. opus)

I can change back though if you prefer

app/dl_formats.py
asuyou: The way I've done it is so that it will choose the best format possible and then in post, will merge video/audio and convert to what ever format is needed.

I originally did this as some videos might not be available in mp4 formats (if I understand yt_dlp correctly, which I probably don't)

I can convert back to the way it was done originally if you prefer

alexta69:
> > Right now when "Any" is selected in Format, it won't let you set the resolution, this needs to be addressed, can you enable resolution selection when Any is selected?
> 
> I'll work on that
> 

There are still some bugs in the UI -- this one still remains, and also when you select a resolution while under "MP4", and then select "MP3", the quality selection becomes empty, while it should switch to "Best"...

> > > Largest change is that best now downloads best possible quality instead of best in a specific file type which can lead to some large files...
> 
> > Not sure what you mean.
> 
> (I might not of understood yt_dlp properly) What I thought was happening originally was that it was downloading the best file with the mp4 extension, whilst this one downloads the best possible file type (e.g. opus)
> 
> I can change back though if you prefer

I suppose that if you want MP4, you have to download MP4, otherwise you'd need transcoding, and we definitely don't want to do that... or perhaps the MP4 will serve as a container format, so an MP4 file would actually contain video that's not MPEG4. which I also think is not what people want when they select MP4. so I think the previous setup was OK.

Again, thanks for your contribution and your effort!

asuyou:
Sorry for taking a while...

The `any` bug should be fixed but I think there is a better way to do it. I did it the way I did to prevent needing to iterate over each time a new item was added.

I changed back to the original format string. I see what I was doing was a bit... dodgy...

Please let me know if you notice any major bugs or if you want me to re-implement something.

Thank you for bearing with me!

app/dl_formats.py
```diff
@@ -0,0 +1,87 @@
...
+
+    if format == "mp4":
+        # video {res} {vfmt} + audio {afmt} {res} {vfmt}
+        vfmt, afmt = "[ext=mp4]", "[ext=m4a]"
+        vres = _get_video_res(quality)
+        combo = vres + vfmt
+        final_fmt = f"bestvideo{combo}+bestaudio{afmt}/best{combo}"
+    elif format == "mp3":
+        final_fmt = _get_audio_fmt(quality)
+    else:
```
alexta69, line 84: does this leave the "any" format unhandled?

ui/src/app/formats.ts
```diff
@@ -0,0 +1,58 @@
...
+];
+
+export const fillQualities = (formats: Format[]): Format[] => {
+  let allQualities: Quality[] = [];
+  formats.forEach((fmt) => {
+    fmt.qualities = fmt.qualities.map((ql) => ({ ...ql, fmt: fmt.id }));
+    allQualities = allQualities.concat(fmt.qualities);
+  });
+
+  formats.find((format) => format.id === 'any').qualities = allQualities;
```
alexta69, line 50: Right now if "Any" format is selected, then the qualities contain both MP3 and MP4 qualities. This is not what we want.
What we want is when Any is selected, to have the same qualities (i.e. video resolutions) as when MP4 is selected. And it should be named "Best".
The audio one can also be named "Best" instead of "Best MP3", since it shows when "MP3" is in the format, so "Best" in the quality is enough..

asuyou: Sorry. I'm not sure I follow...
Shouldn't "Any" show all possible qualities, not just for a specific format?
I have is say "Best MP3" and "Best MP4" so the user knows which format it originates from.

It then passes down the original format as the called one, e.g. if format was "any" and text was "Best MP3" it would be the equivalent of selecting "MP3" as the format, and selecting "Best" as the quality.  

Are you saying that you want any to just show the same things as selecting "MP4"?

Or do you mean that "Any" should not target a specific file type, and would allow file types like "mkv" but with the same qualities available as "mp4"?

If it is the second option, would you prefer to leave quality options as are but not force a specific video format, or make it  only show video options?

app/dl_formats.py
asuyou: Currently, as is, "any" should not be passed down as a format as "Any" just shows all possible options from all other sources.
Thread below explains a bit more.

ui/src/app/formats.ts
alexta69: "Any", until now, meant "any format", while "MP4" forced MPEG4. So you would select the maximum resolution you want, and in the format you'd specify whether you want specifically MPEG4 or "Any" if it doesn't matter.
Now that we're adding MP3, we have to show audio qualities under quality, but I don't want the video options affected by that, so the above should keep working the same way as before.

There is a bit of a mixup here with regards to video vs. audio downloads. Perhaps we should think of a better way to organize all this. I'll think a bit about it. Let me know if you have ideas too. But we definitely want to still allow just selecting "best" and "any" and have it download whatever's available. We have to remember that it's possible to download from many sites beside YouTube here, and for other sites which don't have such quality/format availability, selecting "best" and leaving it to the site to determine what we'll be getting is the only working option.

asuyou: Ahh I see where I got mixed up...
I thought "Any" just meant any format and quality from the ones available

Implementing "any" like that would simplify everything significantly, as we don't need to add and find the format the quality came from.

I'll change the code to move the any option back to working like that.

I'll also see if I can think of a better way to organize

asuyou: > Implementing "any" like that would simplify everything significantly, as we don't need to add and find the format the quality came from.

To clarify, I mean implementing "any" the way where it ignores the format is easier.
The way I was doing was more complicated as it required merging all the formats.

Before I work on fixing this, would you prefer to keep a audio quality in "any" similar to video quality?
Or is it better if we removed it?

Thank you for clarifying and sorry for the earlier mix up!

asuyou:
Just a quick question, do you think we should use UID's for the ID and a separate value, or should we use an ID like "best"  like current master branch?

asuyou:
Sorry it took a while, I was busy last few days.
This should make any work like the original
I kept the format used for storing IDs as I wasn't sure if you wanted that reverted back

Please let me know if you notice a bug or want me to change anything

Sorry for letting this pull request get so long due to me not understanding

alexta69:
Thanks for the changes! It already looks great.
A couple small nitpicks, still. Sorry for dwelling on those details, but let's make it perfect while we're at it :)
Regarding the storing of the IDs, I think we can use the value directly (480, 720 etc.) instead of IDs. Unless I'm missing something.. it doesn't seem to me that the IDs add any value, correct me if I'm wrong.
Also, if I switch the format from Any to MP4 and back, the quality resets. It would make sense to me that it'd stay at 720, for example, if that's what I had selected, because both are video formats. Maybe it happens due to the IDs actually, and if we revert to using values directly, it won't happen. Not sure.
Another very small thing -- the video qualities are sorted from best down to worst, and the MP3 bitrates are sorted the other way around. Perhaps we can reverse them to follow the same logic.
Another suggestion: in dl_formats.py, there are the private functions that start with `_`. They are each only used once though, so they're not really needed, their content can be put at the point they're invoked. I understand the reasoning behind separating them (shorter functions), but it makes the logic harder to follow for me -- you have to jump back and forth in the file to actually follow the conditions and their results. Maybe integrating the code directly will be easier to follow. WDYT?

asuyou:
Thanks for the feedback! I'll try and get the changes done by the weekend.

> Thanks for the changes! It already looks great.
A couple small nitpicks, still. Sorry for dwelling on those details, but let's make it perfect while we're at it :)
> Regarding the storing of the IDs, I think we can use the value directly (480, 720 etc.) instead of IDs. Unless I'm missing something.. it doesn't seem to me that the IDs add any value, correct me if I'm wrong.

Yep, changing it back to that should be fine

> Also, if I switch the format from Any to MP4 and back, the quality resets. It would make sense to me that it'd stay at 720, for example, if that's what I had selected, because both are video formats. Maybe it happens due to the IDs actually, and if we revert to using values directly, it won't happen. Not sure.

Yeah I think that would make sense and work. Worst case scenario, I think just checking if the previous ID is in current we can just switch to that.

My only major concern would be that this might break auto-switching to best, but I'm not sure, so I'll play around and see if I can get it working with the value as ID system or the UID system.

> Another very small thing -- the video qualities are sorted from best down to worst, and the MP3 bitrates are sorted the other way around. Perhaps we can reverse them to follow the same logic.

No problem, will change.

> Another suggestion: in dl_formats.py, there are the private functions that start with `_`. They are each only used once though, so they're not really needed, their content can be put at the point they're invoked. I understand the reasoning behind separating them (shorter functions), but it makes the logic harder to follow for me -- you have to jump back and forth in the file to actually follow the conditions and their results. Maybe integrating the code directly will be easier to follow. WDYT?

I was separating in case someone in the future wanted to add another format and so they wouldn't need to rewrite, but you make a good point about how it's only used once so it should be fine to be in the same block.
Will change to make inside each block.

Once it's all settled, would it be a good idea to squash all the changes into one commit to not clog up master with so many commits?

alexta69:
> Thanks for the feedback! I'll try and get the changes done by the weekend.

Thanks for the perseverance, and for your contribution!

> Once it's all settled, would it be a good idea to squash all the changes into one commit to not clog up master with so many commits?

I think it's OK to merge the commits without squashing, we might as well have this history in there.

asuyou:
I changed quality back to being stored as the ID as it's simpler.
This should work but let me know if you find any bugs or notice anything else that could be optimized.

alexta69:
@asuyou I committed some additional changes (simplified format handling, at the expense of some error checking, but I feel it's a good compromise as it simplifies code readability a great deal), and I merged the PR.
Thanks again for your contribution and all your work and re-work! :)

PR #86, opened 2021-12-01
Try downloading metadata/thumbnails for mp3 files automatically

omeryagmurlu:
Not every audio container has support for thumbnails, so I put it under mp3.

Also, the `ytdl_opts["postprocessors"]` was growing with every download (mutating the parameter is bad), so I fixed that bug too.

### DIFF
--- a/app/dl_formats.py
+++ b/app/dl_formats.py
@@ -1,3 +1,5 @@
+import copy
+
 def get_format(format: str, quality: str) -> str:
     """
     Returns format for download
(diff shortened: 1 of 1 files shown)


alexta69:
Thanks for the contribution!
What happens if there are no thumbnails for a certain download, will it not fail in that case?

omeryagmurlu:
I've just tested it with this video here (http://camendesign.com/code/video_for_everybody/test.html), and downloading mp3 works even when there isn't a thumbnail

asuyou:
Ah thanks for fixing this! Didn't realize the bug. I was going to write a new PR for thumbnails, and just saw this.

Thanks!

Issue #81, opened 2021-11-16
dark mode 


xxcriticxx:
any way to add dark mode to metube?

alexta69:
I don't plan on adding it myself, but if anyone wants to submit a PR, I don't mind merging it

PR #87, opened 2021-12-11
add a simple dark theme with OS theme

PikuZheng:
It will invert the color to make it look dark if your OS selects dark mode. (eg. Windows 10 Settings - Personalization - Colors - App Mode - Dark
#81


alexta69:
@PikuZheng thanks!

Issue #45, opened 2021-08-25
download folder by site 


binarygeek119:
i would like a option to beable to simi sort by site

if i download for youtube.com it goes into /downloads/youtube
if i download for imdb.com it goes into /downloads/imdb

alexta69:
Actually this is already possible by setting the OUTPUT_TEMPLATE variable. Set it to `%(extractor)s/%(title)s.%(ext)s` and you'll get exactly that.

PR #88, opened 2021-12-12
Add download link to downloaded file

### LINKED ISSUE #26: Please add link to download from downloads folder
After downloading video to downloads folder. Please add a URL in the UI to download/delete same video from downloads folder. So we only need webbrowser to do this things and no need to share manually go to downloads folder each time. thanks

kleest:
This adds a simple download link that points to the downloaded file.

Note: This makes all files in the download directory (and its
sub-directory) available to any user.

Closes #26

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -30,6 +30,7 @@ class DownloadInfo:
         self.quality = quality
         self.format = format
         self.status = self.msg = self.percent = self.speed = self.eta = None
+        self.filename = None
 
 class Download:
     manager = None
@@ -60,6 +61,9 @@ class Download:
                     'speed',
                     'eta',
                 )})
+            def put_status_postprocessor(d):
+                if d['postprocessor'] == 'MoveFiles' and d['status'] == 'finished':
+                    self.status_queue.put({'status': 'finished', 'filename': d['info_dict']['filepath']})
             ret = yt_dlp.YoutubeDL(params={
                 'quiet': True,
                 'no_color': True,
@@ -69,6 +73,7 @@ class Download:
                 'cachedir': False,
                 'socket_timeout': 30,
                 'progress_hooks': [put_status],
+                'postprocessor_hooks': [put_status_postprocessor],
                 **self.ytdl_opts,
             }).download([self.info.url])
             self.status_queue.put({'status': 'finished' if ret == 0 else 'error'})
@@ -113,6 +118,8 @@ class Download:
             if status is None:
                 return
             self.tmpfilename = status.get('tmpfilename')
+            if 'filename' in status:
+                self.info.filename = os.path.relpath(status.get('filename'))
             self.info.status = status['status']
             self.info.msg = status.get('msg')
             if 'downloaded_bytes' in status:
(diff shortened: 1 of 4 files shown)


app/ytdl.py
```diff
@@ -113,14 +118,19 @@ async def update_status(self):
             if status is None:
                 return
             self.tmpfilename = status.get('tmpfilename')
+            if 'filename' in status:
+                self.info.filename = os.path.basename(status.get('filename'))
             self.info.status = status['status']
-            self.info.msg = status.get('msg')
+            if 'msg' in status:
```
alexta69, line 124: why do you need the condition? status.get('msg') will return None if 'msg' doesn't exist in status.
(same goes for speed and eta below)

app/ytdl.py
```diff
@@ -113,14 +118,19 @@ async def update_status(self):
             if status is None:
                 return
             self.tmpfilename = status.get('tmpfilename')
+            if 'filename' in status:
+                self.info.filename = os.path.basename(status.get('filename'))
```
alexta69, line 122: sometimes the OUTPUT_TEMPLATE might be set in such a way that downloads are put into directories (see for example here: https://github.com/alexta69/metube/issues/45), in which case this code won't give the right result...

ui/src/app/app.component.html
```diff
@@ -112,7 +112,7 @@
             <fa-icon *ngIf="download.value.status == 'finished'" [icon]="faCheckCircle" style="color: green;"></fa-icon>
             <fa-icon *ngIf="download.value.status == 'error'" [icon]="faTimesCircle" style="color: red;"></fa-icon>
           </div>
-          <span ngbTooltip="{{download.value.msg}}">{{ download.value.title }}</span>
+          <span ngbTooltip="{{download.value.msg}}"><a href="download/{{download.value.filename}}">{{ download.value.title }}</a></span>
```
alexta69, line 115: perhaps it's worth it skipping link creation in case the filename property didn't come in (after all its setting is under an "if" in the code in ytdl.py, so it can conceivably not be set in certain circumstances)

app/ytdl.py
kleest: I thought a 'None' would lead to problems in the UI, because the filename is always sent after download has finished (filename is from the postprocessing step), so the former "good" values of 'msg', 'speed', and 'eta' would be overwritten with 'None'. However, at first glance, it does not cause issues; 'speed' and 'eta' are not required anymore after download has finished.

app/ytdl.py
kleest: Didn't think of that. Good point.

ui/src/app/app.component.html
kleest: I will adjust this. However, I am not experienced with Angular, so feel free to comment on the upcoming changes. :)

app/ytdl.py
alexta69: I see your point, it makes sense. So yeah, speed and eta aren't required, but maybe it's worth keeping msg after all? I haven't thought of the case you're talking about (values being rewritten to None when the download finishes).

app/ytdl.py
alexta69: Do you think we can handle it in some way so that the feature works in these cases too?

kleest: Already pushed an update that uses relpath. Tested with OUTPUT_TEMPLATE from the issue you referenced.

app/ytdl.py
kleest: What does msg contain? If it's only an error message if download/conversion/... fails, then it shouldn't be overwritten since postprocessing should only occur if previous steps are successful.

app/ytdl.py
alexta69: I see now. It's best to push as additional commits, because then it's easier to see what changed since last time. But doesn't matter right now.

alexta69:
Thanks for your contribution!

PR #90, opened 2021-12-16
Made theme toggleable

asuyou:
Uses pull #87  as a base. LMK if you want me to change code style as I split things out a bit

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -23,6 +23,7 @@
         "@fortawesome/free-solid-svg-icons": "^5.15.4",
         "@ng-bootstrap/ng-bootstrap": "^10.0.0",
         "bootstrap": "^4.5.0",
+        "bootstrap-icons": "^1.7.2",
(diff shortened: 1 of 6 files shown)


alexta69:
It's a good idea, thanks. The only thing is the switcher UI element. If we're doing this, let's make it look beautiful too! I'm talking about how it looks here: https://ryanfeigenbaum.com/dark-mode/ and here: https://mui.com/ (which is basically the same) -- the moon / sun button on the right-hand side of the bar at the top. Do you think you can modify the PR to implement that?

alexta69:
I've made some changes, specifically used the icon from fontawesome (since we already use icons from there and we have this dependency) and removed the dependency on bootstrap icons, and moved the toggle switch to the right of the screen.

Thanks @asuyou for the contribution!

asuyou:
Ah got it. Thanks for fixing that!
I didn't realize we were using those icons, my bad...

PR #106, opened 2022-01-15
Queue persistence

Erazor2:
During the past days I experienced some power outages. This lead to a loss of my download queue.

With this PR I'm trying to add persistence to the download-queue and completed-list.

Only the DownloadInfo-Object is persisted.
That allows to display Title and URL for the completed-list.
For the download-queue the persisted items are re-added to the DownloadQueue using url, quality and format on restart.
During my Tests resume of aborted downloads works seamlessly.

A second volume "/queue" was added to the Dockerfile for the queue-files.

I'm by far not python programmer so feel free to modify or tell me about necessary changes.


alexta69:
Thanks for the contribution. What's missing is an environment variable to set the path, similar to DOWNLOAD_DIR. Also, I wouldn't name the directory "queue", perhaps "state", so that other state can be stored there in the future too.. Could you make those changes?
Thanks again

Erazor2:
Sure thing, I hope it looks OK like this.

Renamed directory from queue to state.
Added Env-Variable for the directory.
Added Env-Variable to Readme.

alexta69:
Sorry for not handling it for so long. The reason is that there's one thing I'm not quite happy about, and that is the persistency doesn't keep the order... The OrderedDict keeps the order of the entries, but if it's loaded from the shelve, the order is lost. I'm still trying to think how it can be solved elegantly. If you have any ideas, let me know...

Erazor2:
Order wasn't really a concern for me but I can see why you would prefer to keep things in the initial order.

I added a timestamp to the DownloadInfo which ist used for sorting.
Everything else would require a more complex queue-management (backed by a db not storing the Objects but rather real queue-information)

alexta69:
Have you tested it? I'm having a hard time making it work. After stopping MeTube while there were some items in the queue, and then loading it again, and reloading the browser window, I'm getting this in VSCode:
image
I'm still investigating it. In any case, it did seem strange to me that the downloads are loaded only when the UI is loaded. Wouldn't it make more sense to load them when MeTube boots the next time?

Erazor2:
On my end it works.
Did you delete the old database files? The ones from previous commit will be incompatible.

About loading the stored entries... You'r absolutely right, the downloads should be rescheduled when metube starts, but I couldn't get it working. As I said, I'm no python programmer and not familiar with the python asyncio

alexta69:
I've reworked the code a bit, and made it load the saved downloads at startup.
Thanks a lot for your contribution!

Issue #107, opened 2022-01-16
Feature request: Download playlists to a subdirectory


matt-bartel:
With yt-dlp it's possible to download a playlist to a subdirectory by providing an output template (example: `%(playlist_title)s/%(title)s.%(ext)s`). Unfortunately that isn't possible with metube since it gets the playlist info then passes each video on to yt-dlp rather than the playlist url, meaning yt-dlp isn't operating in the context of a playlist and the playlist fields can't be used in the output format.

I fixed this for myself by adding a new config option (`PLAYLIST_DIR_TEMPLATE`) but haven't opened a PR for this since it feels wrong to me. Rather than using yt-dlp's output formatting it takes a python format string. [Here's my branch](https://github.com/matt-bartel/metube/tree/playlist-subfolder) if you're interested.

alexta69:
Yeah, I see. Indeed it's not ideal, though I can't think of a better solution at this point... I'll leave it as is for now, and hopefully someday someone will come up with one :)
Thanks!

Issue #112, opened 2022-01-30
Renaming Downloaded Videos 


Hussienfahmy:
Can you please add a feature to rename the videos with the same order as it in the playlist 
for example
1 - Video_1.mp4
2 - Video_2.mp4
3 - Video_3.mp4

alexta69:
If someone takes this ticket, please talk to me in advance so we can decide how to do it.

PikuZheng:
try to change OUTPUT_TEMPLATE in ENV. `%(playlist_index)` is the index of the video in the playlist

alexta69:
@PikuZheng this won't work, most likely, for the reasons mentioned in #107 ...

PR #120, opened 2022-02-17
add playlist parameters

almeidapaulopt:
Passes playlist parameters to each video 

resolves:
https://github.com/alexta69/metube/issues/107
https://github.com/alexta69/metube/issues/112

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -215,10 +215,9 @@ class DownloadQueue:
             for index, etr in enumerate(entries, start=1):
                 etr["playlist"] = entry["id"]
                 etr["playlist_index"] = '{{0:0{0:d}d}}'.format(playlist_index_digits).format(index)
-                etr["playlist_id"] = entry["id"]
-                etr["playlist_title"] = entry["title"]
-                etr["playlist_uploader"] = entry["uploader"]
-                etr["playlist_uploader_id"] = entry["uploader_id"]
+                for property in ("id", "title", "uploader", "uploader_id"):
+                    if property in entry:
+                        etr[f"playlist_{property}"] = entry[property]
                 results.append(await self.__add_entry(etr, quality, format, already))
             if any(res['status'] == 'error' for res in results):
                 return {'status': 'error', 'msg': ', '.join(res['msg'] for res in results if res['status'] == 'error' and 'msg' in res)}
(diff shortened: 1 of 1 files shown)


app/ytdl.py
```diff
@@ -210,8 +210,15 @@ async def __add_entry(self, entry, quality, format, already):
...
+            playlist_index_digits = len(str(len(entries)))
             results = []
-            for etr in entries:
+            for index, etr in enumerate(entries, start=1):
+                etr["playlist"] = entry["id"]
+                etr["playlist_index"] = '{{0:0{0:d}d}}'.format(playlist_index_digits).format(index)
+                etr["playlist_id"] = entry["id"]
+                etr["playlist_title"] = entry["title"]
+                etr["playlist_uploader"] = entry["uploader"]
+                etr["playlist_uploader_id"] = entry["uploader_id"]
```
alexta69, line 221: can we be certain that the "uploader" and "uploader_id" properties will always exist in those entries? yt-dlp supports a lot of sites, maybe the extractors for some of them don't fill those fields... perhaps something like this can be done:
```suggestion
                for property in ("id", "title", "uploader", "uploader_id"):
                    if property in entry:
                        etr[f"playlist_{property}"] = entry[property]
```

app/ytdl.py
```diff
@@ -220,7 +227,15 @@ async def __add_entry(self, entry, quality, format, already):
...
                 dldirectory = self.config.DOWNLOAD_DIR if (quality != 'audio' and format != 'mp3') else self.config.AUDIO_DOWNLOAD_DIR
-                self.queue.put(Download(dldirectory, self.config.OUTPUT_TEMPLATE, quality, format, self.config.YTDL_OPTIONS, dl))
+                output = self.config.OUTPUT_TEMPLATE
+                if "playlist" in entry:
+                    output = output.replace("%(playlist_index)s",entry["playlist_index"])
+                    output = output.replace("%(playlist_id)s",entry["playlist_id"])
+                    output = output.replace("%(playlist_title)s",entry["playlist_title"])
+                    output = output.replace("%(playlist_uploader)s",str(entry["playlist_uploader"]))
+                    output = output.replace("%(playlist_uploader_id)s",str(entry["playlist_uploader_id"]))
+                    output = output.replace("%(playlist)s",entry["playlist"])
```
alexta69, line 237: here we can write something generic that copes any playlist property that was supplied, so the code is more resilient to such changes:
```suggestion
                for property, value in entry.items():
                    if property.startswith("playlist"):
                        output = output.replace(f"%({property})s", value
```

alexta69 (changes requested):

almeidapaulopt:
> 

Yes, I've tested with some yt playlists and a single videos. Worked fine.

alexta69:
Thanks a lot for your contribution!

PR #146, opened 2022-05-29
use gosu and usermod to ensure data permissions

VergilGao:
Hi, I've been using metube for a while now, thanks for your great work making it easier for me to use youtube-dl

I noticed that the current handling of file permissions is to use `--user` of docker. However, for most  nas system, the setting of `--user` is not convenient. This can be solved by passing the UID and GID as environment variables into the image.

this commit works, but left some document job to do.


alexta69:
Hi!
Thanks for the contribution!
A couple of questions:
1. It seems that https://github.com/ncopa/su-exec does the same thing, and it's already available in alpine, so using it instead of goexec will greatly simplify the build. I think it's worth checking..
2. A question arises, what happens to all the existing installations which already have the `user:` and `group:` settings in the docker-compose file. Will there be continuity for them, or will it break? I'd say it's worth checking and adding logic, if necessary, to avoid breaking in these cases.

VergilGao:
ok
i will add a check if the docker not run as root user.

alexta69:
> ok
> 
> > It seems that https://github.com/ncopa/su-exec does the same thing, and it's already available in alpine, so using it instead of goexec will greatly simplify the build. I think it's worth checking..
> 
> I think `gosu` is better, especially it is well maintained and already included in distributions like debian

There isn't much to maintain there though, it's pretty minimal and straightforward functionality.. and it's in the main distribution of alpine, which means that they're making sure it keeps working. We're not using Debian in this project so the fact that it's there doesn't really help us. Using gosu means adding another stage to the docker build, lengthening it, complicating it, and making it more fragile, without any actual practical advantage that I can think of.

> > A question arises, what happens to all the existing installations which already have the user: and group: settings in the docker-compose file. Will there be continuity for them, or will it break? I'd say it's worth checking and adding logic, if necessary, to avoid breaking in these cases.
> 
> I will add a check if the docker not run as root user.

Thanks!

VergilGao:
> There isn't much to maintain there though, it's pretty minimal and straightforward functionality.. and it's in the main distribution of alpine, which means that they're making sure it keeps working. We're not using Debian in this project so the fact that it's there doesn't really help us. Using gosu means adding another stage to the docker build, lengthening it, complicating it, and making it more fragile, without any actual practical advantage that I can think of.

OK, you persuaded me.
I will finish my job this weekend.

VergilGao:
@alexta69 
It looks like I've done my work, but I'm not a native English speaker, maybe the documentation part needs some polish.

docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+echo "You are running with user `id -u`:`id -g`"
+
+if [ `id -u` -eq 0 ] && [ `id -g` -eq 0 ]; then
+    echo "Running in New Mode"
+    if [ "${UID}" -eq 0 ]; then
+        echo "Waring, it is not recommended to run as root user, please check if you have set the UID environment variable"
```
sfragata, line 8: typo `Warning`

sfragata (changes requested):

sfragata (approved):

alexta69:
Sorry for taking so long. I refactored some things, and pushed it to master. I hope the changes are all right by you. Thanks for your contribution!

PR #149, opened 2022-06-06
Bugfix: Use paths for providing path information instead of implicit path in output template

georgekav2:
In most cases, using path inside `outtmpl` would be enough. However, now with the different post processors, like the `split-chapters` / `FFmpegSplitChapters` , the `outtmpl` is not shared between them; actually there are various types defined in `OUTTMPL_TYPES`. 

By providing the path on the default only `outtmpl` creates issues like: [yt-dlp issue](https://github.com/yt-dlp/yt-dlp/issues/3979). For metube the issue is that the download_dir path is not respected and we get permission denials. 

To avoid that, we need to explicitly pass the paths dictionary which is respected by all post processors for the output of the files they process even without defining their proper `outtmpl` for their type. 

This PR  is also defining in addition a `OUTPUT_TEMPLATE_CHAPTER` that is passed to yt-dlp for formatting the chapter type of output files when used in the similar template with the `OUTPUT_TEMPLATE` of metube.


alexta69:
Sorry for the delay in merging, and thanks for your contribution!

PR #172, opened 2022-08-25
Added a couple env variables

unref:
Hi, this commit add ability to set port, metube listen address and application directory as env variables.


alexta69:
Can you explain the use case? MeTube is meant to run under Docker. Under which circumstances would anyone want to change these values?

unref:
I see that MeTube is meant run under docker and that's awesome. In my case, I use MeTube in pair with the LXC container, I think it can be useful in a similar case or when running on bare hardware. Ability to set host ADDRESS and PORT in this case very necessary to config and run as old style init script. For instance under alpine linux driven by OpenRC init system.

alexta69:
I understand. In that case I prefer to leave them undocumented, since I think they're only going to confuse the majority of the users (who are using it with Docker). Perhaps if you can just remove the documentation from the README, and then we can push it.
Thanks!

wy580477:
> Can you explain the use case? MeTube is meant to run under Docker. Under which circumstances would anyone want to change these values?

I hope I‘m not being rude to jump in this conversation. 

I mostly perfer host network for my containers since bridge networks gave me troubles from time to time. Being able to set port and listening addr would be very useful for host network.

Thx for your awesome work.

unref:
Hi, can you merge this feature. Removed notice about it this from `README.md`. Thanks

PR #177, opened 2022-08-30
Custom download folders (fixes #129)

### LINKED ISSUE #129: Ability to download to a different folder
Hi

I'm aware you can use output templates to direct the downloads to different folders but there's no way to do it with individual vidoes. Say for example I want to download a video that should be in a 'Kids' folder and then download one to 'Music Videos' it's not possible through the frontend

Thanks

jwoglom:
Adds a "Download Folder" option in the UI, hidden behind a split-button dropdown.

image

The menu doesn't appear unless the caret is clicked.

image

The Download Folder option is a [ng-select component](https://github.com/ng-select/ng-select), which allows selecting an existing folder in the filesystem like a standard select or via a fuzzy search. This is gated behind the `CUSTOM_DIRS` configuration field, which defaults to true.

image

image

An optional `CREATE_DIRS` configuration field was added, disabled by default, which when set allows for directories to be created via this UI. For example, to create a new folder in which to save a video, click the "Create directory" option or press enter after typing the folder path:

image

image

If the `AUDIO_DOWNLOAD_DIR` option is used, then when either the Quality selector is set to "Audio Only" or the Format selector is set to "MP3" (matching existing logic on the backend), then the folders displayed change to that of the audio download directory instead of the `DOWNLOAD_DIR`.

image

@alexta69 I hope this alleviates some of the concerns around complicating the UI while also allowing for some pretty powerful new behaviors via this feature.

A docker image with these changes for testing should be (soon) available at https://hub.docker.com/r/jwoglom/metube


app/main.py
```diff
@@ -16,11 +17,13 @@ class Config:
     _DEFAULTS = {
         'DOWNLOAD_DIR': '.',
         'AUDIO_DOWNLOAD_DIR': '%%DOWNLOAD_DIR',
+        'CUSTOM_DIRS': 'true',
+        'CREATE_DIRS': 'false',
```
alexta69, line 21: I would remove CUSTOM_DIRS altogether (I don't see a use case where it wouldn't be true), and would set CREATE_DIRS by default to true. (And perhaps change the name to be more descriptive, since CREATE_DIRS isn't telling enough, maybe ALLOW_DOWNLOAD_DIR_CREATE? Or something similar if you can think of something shorter :) )

app/main.py
```diff
@@ -16,11 +17,13 @@ class Config:
...
         'DOWNLOAD_DIR': '.',
         'AUDIO_DOWNLOAD_DIR': '%%DOWNLOAD_DIR',
+        'CUSTOM_DIRS': 'true',
+        'CREATE_DIRS': 'false',
         'STATE_DIR': '.',
         'URL_PREFIX': '',
         'OUTPUT_TEMPLATE': '%(title)s.%(ext)s',
         'OUTPUT_TEMPLATE_CHAPTER': '%(title)s - %(section_number)s %(section_title)s.%(ext)s',
-        'YTDL_OPTIONS': '{}',
+        'YTDL_OPTIONS': '{}'
```
alexta69, line 26: why the change? (the comma is there to make the diff of adding a new variable at the end a bit shorter)

alexta69:
Hi! Thanks so much for the contribution! Sorry I only got to it now, I was on an extended vacation.
I posted a couple of small things in the code, take a look, and then we can push it in.
It's a great addition.

PR #178, opened 2022-09-01
Added download button to the ui

cavazzatommaso:
For personal needs, I've added the download button to the completed downloads


alexta69:
The video name itself is already a link for the download, why would you need a separate download button?

cavazzatommaso:
If you are hosting the application locally, if you download a video, it will be saved on your server not your device. Adding this button, you don't need to open the video to a new tab, but instead you just click the download button and it will download to your machine

ui/src/app/app.component.html
```diff
@@ -1,6 +1,6 @@
 <nav class="navbar navbar-expand-md navbar-dark bg-dark">
   <div class="container-fluid">
-    <a class="navbar-brand" href="#">MeTube</a>
+    <a class="navbar-brand" href="#">Youtube downloader</a>
```
alexta69, line 3: why this change?

ui/src/app/app.component.html
```diff
@@ -124,6 +125,9 @@
         <td>
           <button *ngIf="download.value.status == 'error'" type="button" class="btn btn-link" (click)="retryDownload(download.key, download.value.url, download.value.quality)"><fa-icon [icon]="faRedoAlt"></fa-icon></button>
         </td>
+        <td>
+          <a *ngIf="!!download.value.filename; else noDownloadLink" href="download/{{download.value.filename | encodeURIComponent}}" download target="_blank"><fa-icon [icon]="faDownload"></fa-icon></a>
```
alexta69, line 129: do you need the `target` property when you have a `download` link? it's not going to open the file, just download it, so the target seems irrelevant, unless I'm missing something

alexta69:
OK, I understand. I added a couple of comments on the commit, check them out
Thanks!

ui/src/app/app.component.html
cavazzatommaso: Ops sorry my bad, this can be deleted. I've used it in developer to see changes in UI and forget to delete

ui/src/app/app.component.html
cavazzatommaso: Yep target is irrelevent. Only the keyword "download"

kblock1, line 129: This download link doesn't work, the folder "download/" is not accurate.  This works better:

<a *ngIf="!!download.value.filename; else noDownloadLink" href="{{buildDownloadLink(download.value)}}" download><fa-icon [icon]="faDownload"></fa-icon></a>

PR #216, opened 2023-02-03
unique downloads of identically named videos

### LINKED ISSUE #124: Downloads with the same title / req: manually enter title
I have links to m3u8 playlists (for a stream) that all give the same title "playlist" (and same output file name) - this means I can only enter one at a time to download list. Presumably title is used for the index of the playlist? and the subsequent ones don't get added because it's the same.

I've set OUTPUT_TEMPLATE   to  %(epoch-3600>%H-%M-%S)s-%(title)s.%(ext)s  which does prefix the downloaded file with a unique timestamp, but still doesn't allow multiple entries in the queue with the same title.

I'm not sure what the best workaround would be, detect and append (1) (2) etc like windows file manager? allow overriding the title with an optional field (awkward UI)?

Currently I have to wait for one to finish before adding another which kind of defeats the purpose.

any help appreciated!

### LINKED ISSUE #139: If you download two videos with the same name, only the first video will be downloaded.
Hello, I'm running MeTube in a Docker container and have recently experienced this issue. Has anybody have a fix for it? Thanks :)

### LINKED ISSUE #214: Non-unique filenames lead to silent queue addition failures
Some sites/extractors return non-unique filenames (for instance `stream.m3u8`) which are then used to check uniqueness in the queue:

https://github.com/alexta69/metube/blob/9bbd92919a854de1c636281f769481cef3c3cf90/app/ytdl.py#L173

An alternate approach would be to use the url as the queue key, which should be available via `value.info.get('webpage_url') or value.info['url']`. I don't have time to check at the moment but I imagine this change would allow different URLs that map to the same video "id" to be added to the queue at the same time.

Edit: this fix may solve #124 and #139

kaytwo:
closes #214.

also closes #124 and closes #139 as long as you use an `OUTPUT_TEMPLATE` like `%(title)s.%(epoch)s.%(ext)s` to uniquify the downloaded filename.

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -170,7 +170,7 @@ class PersistentQueue:
             return sorted(shelf.items(), key=lambda item: item[1].timestamp)
 
     def put(self, value):
-        key = value.info.id
+        key = value.info.url
         self.dict[key] = value
         with shelve.open(self.path, 'w') as shelf:
             shelf[key] = value.info
--- a/ui/src/app/downloads.service.ts
+++ b/ui/src/app/downloads.service.ts
@@ -52,20 +52,20 @@ export class DownloadsService {
     });
     socket.fromEvent('added').subscribe((strdata: string) => {
       let data: Download = JSON.parse(strdata);
-      this.queue.set(data.id, data);
+      this.queue.set(data.url, data);
       this.queueChanged.next(null);
     });
     socket.fromEvent('updated').subscribe((strdata: string) => {
       let data: Download = JSON.parse(strdata);
-      let dl: Download = this.queue.get(data.id);
+      let dl: Download = this.queue.get(data.url);
       data.checked = dl.checked;
       data.deleting = dl.deleting;
-      this.queue.set(data.id, data);
+      this.queue.set(data.url, data);
     });
     socket.fromEvent('completed').subscribe((strdata: string) => {
       let data: Download = JSON.parse(strdata);
-      this.queue.delete(data.id);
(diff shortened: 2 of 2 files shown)


alexta69:
Hi! Thanks for the PR!
You mention in #214 that the problem is with non-unique filenames, but the `id` is not the filename. The `id` is supposed to be unique, as far as I understand. Are you sure that this change solves the problem? Do you have an example of two different videos with the same `id` (but different URLs)?
And another question. The queues are persisted in files. What happens if the new version (after the change) loads old entries from the queue? I just want to make sure we're not breaking existing installations.

kaytwo:
> but the `id` is not the filename

I didn't look into it super deeply, but I believe that `yt-dlp` uses the filename as the default id if there isn't one created by the extractor - when downloading raw mp4's from my testing server (see below) that's the behavior I saw.

I tested it using this script

```bash
#! /opt/homebrew/bin/lighttpd -Df

server.document-root = var.CWD
server.kbytes-per-second = 1000
server.port = 8080
debug.log-request-handling = "enable" 

mimetype.assign = (
  ".mp4" => "video/mp4",
  ".html" => "text/html", 
  ".txt" => "text/plain",
  ".jpg" => "image/jpeg",
  ".png" => "image/png" 
)
```

With a `test.mp4` and `subdir/test.mp4` in the cwd. Both queues were able to be reloaded on restart, and everything else worked as expected.

alexta69:
Thanks again for your contribution!

PR #221, opened 2023-02-18
Add m4a support and explicit audio format

georgekav2:
This PR add m4a audio format support that is native to many ytb videos and could avoid transcoding. 
Also replaces the "Any" format, with explicit format selector which will also drive the quality selector.  

image
image
image

### DIFF
--- a/app/dl_formats.py
+++ b/app/dl_formats.py
@@ -23,14 +23,13 @@ def get_format(format: str, quality: str) -> str:
         # Quality is irrelevant in this case since we skip the download
         return "bestaudio/best"
 
-    if format == "mp3":
+    if format in ("m4a", "mp3"):
         # Audio quality needs to be set post-download, set in opts
         return "bestaudio/best"
 
     if format in ("mp4", "any"):
         if quality == "audio":
             return "bestaudio/best"
-
         # video {res} {vfmt} + audio {afmt} {res} {vfmt}
         vfmt, afmt = ("[ext=mp4]", "[ext=m4a]") if format == "mp4" else ("", "")
         vres = f"[height<={quality}]" if quality != "best" else ""
@@ -60,12 +59,13 @@ def get_opts(format: str, quality: str, ytdl_opts: dict) -> dict:
     if "postprocessors" not in opts:
         opts["postprocessors"] = []
 
-    if format == "mp3":
+    if format in ("m4a", "mp3"):
         opts["postprocessors"].append({
             "key": "FFmpegExtractAudio",
-            "preferredcodec": "mp3",
+            "preferredcodec": format,
             "preferredquality": 0 if quality == "best" else quality,
         })
+
         opts["writethumbnail"] = True
         opts["postprocessors"].append({"key": "FFmpegThumbnailsConvertor", "format": "jpg", "when": "before_dl"})
         opts["postprocessors"].append({"key": "FFmpegMetadata"})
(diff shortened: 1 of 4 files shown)


alexta69:
Why remove "any"? And also, what happens if someone already has "any" in their cookie?

PikuZheng:
I don't think removing "any" is a good idea. Websites other than YouTube may not match the appropriate format or resolution. At this time, "any" can make yt-dlp choose independently to avoid failure.

georgekav2:
Thanks for your feedback, indeed, if Any was in cookies it might fail to load the page!

I did not think the non-existence of any would affect other than Youtube as it seems yt-dlp to make the good choice from existing streams and post-process it to the format you choose. 

However, looking at the broad picture, keeping "Any" as a catch all format to retain maximum compatibility and push the choice task to yt-dlp makes sense. I updated the commits accordingly.

app/dl_formats.py
```diff
@@ -60,12 +57,14 @@ def get_opts(format: str, quality: str, ytdl_opts: dict) -> dict:
...
-    if format == "mp3":
+    if format in ("m4a", "mp3"):
         opts["postprocessors"].append({
             "key": "FFmpegExtractAudio",
-            "preferredcodec": "mp3",
+            "preferredcodec": format,
             "preferredquality": 0 if quality == "best" else quality,
         })
+    
+    if format in ("m4a", "mp3"):
```
alexta69, line 67: why the additional "if" statement? seems like it's identical to the one above, so it's not necessary

app/dl_formats.py
```diff
@@ -23,14 +23,11 @@ def get_format(format: str, quality: str) -> str:
         # Quality is irrelevant in this case since we skip the download
         return "bestaudio/best"
 
-    if format == "mp3":
+    if format in ("m4a", "mp3"):
         # Audio quality needs to be set post-download, set in opts
         return "bestaudio/best"
 
-    if format in ("mp4", "any"):
```
alexta69, line 30: it seems like the "any" case is not handled anymore in this function, and would result in an exception

app/ytdl.py
```diff
@@ -234,7 +234,7 @@ async def __add_entry(self, entry, quality, format, folder, already):
             if not self.queue.exists(entry['id']):
                 dl = DownloadInfo(entry['id'], entry['title'], entry.get('webpage_url') or entry['url'], quality, format, folder)
                 # Keep consistent with frontend
-                base_directory = self.config.DOWNLOAD_DIR if (quality != 'audio' and format != 'mp3') else self.config.AUDIO_DOWNLOAD_DIR
```
alexta69, line 237: I know this code has already been here before, but I'm wondering if the condition makes sense -- why predicate on the format? It makes sense to me that if `quality == "audio"` we should use `AUDIO_DOWNLOAD_DIR`, and that's that

ui/src/app/app.component.ts
```diff
@@ -94,7 +94,7 @@ export class AppComponent implements AfterViewInit {
   }
 
   isAudioType() {
-    return this.quality == 'audio' || this.format == 'mp3';
+    return this.quality == 'audio' || this.format == 'mp3'  || this.format == 'm4a';
```
alexta69, line 97: Same here, perhaps we just need to test for `this.quality == 'audio'`

app/ytdl.py
PikuZheng: Looks like a writing habit problem.  When quality is not audio and format is not mp3, do nothing, otherwise use audio dir.  
That is, when the quality is audio or the format is mp3, use audio dir(⁠ノ⁠｀⁠Д⁠´⁠)⁠ノ⁠彡⁠┻⁠━⁠┻

alexta69: actually I do see the logic. we can select audio both in the quality and in the format. it's a bit of a mess. oh well.

georgekav2: Yes that is why I was intended in the beginning to drop Any that would needed to have Audio selected as quality.

Tuhinakhand (changes requested):
Yes

Issue #212, opened 2023-01-30
Ogg and Wav Audio Formats?


Eoin-ONeill-Yokai:
Hey, this is a really nice tool and I think it's pretty useful as a convenient frontend. I've noticed that there's a list of formats that are modifiable in `/ui/src/app/formats.ts` that would allow me to add Ogg support if I were to build the project from source and host my own docker image.

However, I was wondering if you could add support for these formats for two key reasons:
1) Ogg vorbis format can be a format with better licensing conditions, and thus be a format of choice for many open source projects and applications.
2) Wav format would give some users some leeway to download a file in best-possible quality and reduce the number of re-encodes, preserving audio quality to as-close-to-source as possible. This would be useful when you need to manually split a long-form video into multiple audio files, such as when downloading a whole album.

Thanks for this great tool btw!

PikuZheng:
I think you need to modify format.ts and dl_formats.py to use ffmpeg to convert it to ogg format after downloading. You can submit a pr.

Presjar:
Just wondering if the same method can be used to allow mkv as the resulting file format?

In the [yt-dlp documentation](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/YoutubeDL.py#L183)
'format' can be used, but I'm guessing If I try to use this method via the _YTDL_OPTIONS: Additional options to pass to youtube-dl, in JSON format._ for the docker it will conflict with the Format selection in the webui making that selection useless?

alexta69:
No, there won't be a conflict, whatever you specify in `YTDL_OPTIONS` just overrides what you have in the UI. mkv is a bit different from others because it's a container format, not the encoding format, so it can contain any encoding. So it doesn't belong in the same place. For now, just use `YTDL_OPTIONS`...

alexta69:
Fixed in #230

Eoin-ONeill-Yokai:
Awesome, thank you so much!

I was going to look into it but never found the time.

ChrisKolan:
@Eoin-ONeill-Yokai asked about ogg format but opus was merged instead. Are there plans to add ogg vorbis format?

alexta69:
@ChrisKolan no plans, but feel free to submit a PR

PR #230, opened 2023-03-04
Add support for opus and wav

georgekav2:
Add two more audio formats. Addresses also feature request https://github.com/alexta69/metube/issues/212.

### DIFF
--- a/app/dl_formats.py
+++ b/app/dl_formats.py
@@ -1,5 +1,7 @@
 import copy
 
+AUDIO_FORMATS = ("m4a", "mp3", "opus", "wav")
+
 def get_format(format: str, quality: str) -> str:
     """
     Returns format for download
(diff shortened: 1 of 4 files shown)


app/dl_formats.py
```diff
@@ -23,7 +23,7 @@ def get_format(format: str, quality: str) -> str:
         # Quality is irrelevant in this case since we skip the download
         return "bestaudio/best"
 
-    if format in ("m4a", "mp3"):
+    if format in ("m4a", "mp3", "opus", "wav"):
```
alexta69, line 26: Since there's a lot of repetition of this quartet, perhaps it's worth creating a constant `AUDIO_FORMATS` or something like that, and using it in all those `if`s

georgekav2:
You are right, thanks

alexta69:
Thanks!

georgekav2:
As for the format=Any together with the option for Audio-only in quality should be maintained for compatibility with other streaming sites?

alexta69:
I'm not sure what's the best way to organize it better, and keep this compatibility... It sort of grew organically into this, from all kinds of PRs from different people, it's not a result of a conscious decision.

georgekav2:
Would "Any-Video" or "Any-Audio" in format could make sense? In quality then it will include only quality elements that make sense on the type.

alexta69:
We do have to include some sort of "default" quality though, that would leave it unspecified when calling the actual downloader, so that the decision is left to the downloader, to keep max compatibility.

Issue #234, opened 2023-03-12
[Feauture request] custom name


hellodword:
I'm using metube to download and convert m3u8 to mp4, but the m3u8 resouces always have no file name infomation in the links or the responses, so they're always named to `master.mp4` if the link is end of `master.m3u8`.

It'll be great if metube can set a custom name when adding a new task.

PikuZheng:
with http post method, I can assign a ”video title“.
How did you add the task? using the ui normally?

PikuZheng:
I can submit the name while increasing the download queue https://github.com/PikuZheng/metube/commit/3d8e9c40de19bd8e026e9c6370b6396e20022ad2 
But I don't know how to add it in the ui.  I've been using it for a long time in greasemonkey myself.

alexta69:
Fixed by #236 .

PR #236, opened 2023-03-15
enable custom name

hellodword:
solve #234 

I do not think this pr is good enough, glad if you can review it, I'll try my best to fix, thanks.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -94,7 +94,10 @@ async def add(request):
         raise web.HTTPBadRequest()
     format = post.get('format')
     folder = post.get('folder')
-    status = await dqueue.add(url, quality, format, folder)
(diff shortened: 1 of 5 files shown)


alexta69:
Thanks for the PR! Some notes, perhaps it can be improved.

If you set a custom name, download, and then clear the field and download again, the next download starts with a dot. Apparently the string remains empty but not null, or something like that. Can you try to reproduce and find why this happens?

Also, I'd call it "Custom name prefix" or something like that, because it doesn't replace the whole name.

Also, in the UI, the fields are stuck together very close and it's not too pretty.. Bootstrap has ways to manage it, I think it needs to be improved. Perhaps a "container" div and "row" divs within it, like the whole input section itself..

Thanks again.

alexta69:
Thanks @hellodword !

PR #237, opened 2023-03-18
Create build-ghcr-docker-image.yml

rcarmo:
This builds a docker image and pushes it to the GitHub container registry using the default (repo owner) tokens - but only when a new git tag is pushed.

### DIFF
--- a/.github/workflows/build-ghcr-docker-image.yml
+++ b/.github/workflows/build-ghcr-docker-image.yml
@@ -1,0 +1,27 @@
+name: Build Docker Image
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+      - name: Login to Registry
+        uses: docker/login-action@v2
+        with:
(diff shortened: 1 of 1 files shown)


alexta69:
Looks like it builds the images for a second time, why would we do it twice? Can't we push to ghcr in the main job? And also, why the condition on the tag? We're not using such tags here.

alexta69:
By building twice, I meant that it's repeating the building work that's already done in main.yml (the main pipeline). Of course your file doesn't build twice. And thanks for the suggestion! Perhaps it's a good idea to push the image to ghcr in addition to Docker Hub, there's no harm in that.

What do you mean by "gone from Docker Hub" though? It's there... https://hub.docker.com/r/alexta69/metube

rcarmo:
I use watchtower to update my images automatically, and as of a week or so ago requests for free tier images seem to be throttled, so I am getting errors daily and am reaching out to the developers one by one to suggest they move their images to ghcr :)

alexta69:
I've added pushing to ghcr in the main pipeline, so now images are continuously pushed to both Docker Hub and the GitHub registry. So you're welcome to switch pulling from ghcr.io. Thanks for the suggestion and the PR!

PR #247, opened 2023-04-10
Fix error because of missing cache directory

1RandomDev:
Previously when downloading age restricted video from YouTube you got the error "PermissionError: [Errno 13] Permission denied: '/.cache'" from yt-dlp. This pull request fixes it by creating the necessary directory with the correct permissions in the container.


alexta69:
Thanks! It's been a long-standing issue :)

PR #265, opened 2023-05-09
Temp dir support

### LINKED ISSUE #257: File disappears when setting a temp directoy
I have a SMB Share for storing downloads but its a bit slow, especially if transcoding with ffmpeg is done, so i set up a temp path in the YTDL_OPTIONS. 
Works fine for the downloading and transcoding part, but when it is finished, the File disappears. It does not get moved/copied to the download folder.
I changed the download folder to a local directory to see if it is a problem with cifs/smb, but it also disappears. I cannot see any errors in the log of the docker container.

breakid:
Fixes #257

- Added support for yt-dlp `temp` path
- Formatted with Black
- Updated README to reflect new `TEMP_DIR` setting; linted
- Modified Dockerfile to strip carriage return characters from docker-entrypoint.sh script to fix building the image on Windows
- Added example docker-compose.yml config


Dockerfile
```diff
@@ -10,10 +10,10 @@ FROM python:3.8-alpine
 
 WORKDIR /app
 
-COPY Pipfile* docker-entrypoint.sh ./
```
alexta69, line 13: Why separate the copy? Each command in the dockerfile creates another layer, so this is just adding extra redundant layers.

Dockerfile
```diff
@@ -23,6 +23,11 @@ RUN chmod +x docker-entrypoint.sh && \
     mkdir /.cache && chmod 777 /.cache
 
 COPY favicon ./favicon
+COPY docker-entrypoint.sh .
+
+# Use sed to strip carriage-return characters from the entrypoint script (in case building on Windows)
+RUN sed -i 's/\r$//g' docker-entrypoint.sh && chmod +x docker-entrypoint.sh
```
alexta69, line 29: Same goes for the separate RUN command, better to have all of them in one command.

README.md
```diff
@@ -3,7 +3,7 @@
 ![Build Status](https://github.com/alexta69/metube/actions/workflows/main.yml/badge.svg)
 ![Docker Pulls](https://img.shields.io/docker/pulls/alexta69/metube.svg)
 
-Web GUI for youtube-dl (using the [yt-dlp](https://github.com/yt-dlp/yt-dlp) fork) with playlist support. Allows you to download videos from YouTube and dozens of other sites (https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md).
+Web GUI for youtube-dl (using the [yt-dlp](https://github.com/yt-dlp/yt-dlp) fork) with playlist support. Allows you to download videos from YouTube and dozens of other sites (<https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md>).
```
alexta69, line 6: What do the brackets do?

app/ytdl.py
```diff
@@ -58,40 +64,54 @@ def __init__(self, download_dir, output_template, output_template_chapter, quali
...
-                    'downloaded_bytes',
-                    'speed',
-                    'eta',
-                )})
+                self.status_queue.put(
+                    {
+                        k: v
+                        for k, v in st.items()
+                        if k
+                        in (
```
alexta69, line 74: I think this reformatting is not doing a good service to readability.. Is there any chance you could create a version of this PR without the reformatting?

alexta69 (changes requested):

Dockerfile
breakid: I mostly separated the COPY and RUN commands for development purposes so I didn't have to sit through Docker installing all the packages if/when I changed the docker-entrypoint.sh file. I'm not opposed to merging them back if that's your preference.

Dockerfile
breakid: See above. I'll merge it back in.

README.md
breakid: I think my Markdown linter automatically added those because it was a bare URL. I'll tweak that.

app/ytdl.py
breakid: Sure thing. That was another auto-formatter change; I'll put it back the way it was.

breakid:
I've made another commit with the changes you requested.

app/ytdl.py
alexta69: I commented on this example, but I meant the whole auto-formatter thing. I'd really rather not push the auto-formatting changes at all. Let's just push actual changes in logic

breakid: I misunderstood; my apologies. I've reverted all formatting changes, leaving just the logic changes.

breakid:
I've reverted all formatting changes, leaving just the logic changes.

app/ytdl.py
PikuZheng, line 1: I'm not familiar with python. but it looks like it hasn't changed?

alexta69:
Hey. Thanks and sorry it took so long. I really have too little time to work on this..
I removed docker-compose.yml prior to merging as it might be confusing for users (it's building metube locally, so it's not really for users). I hope that's all right by you.
Thanks again for the contribution!

breakid:
I completely understand; I have the same problem! Glad I could help; thanks for merging my changes.

That's fine. I figured users would just pull the official image and anyone cloning the repo would want to build it locally (plus I had the official image in there commented out so users could switch back and forth), but it was basically straight from the README, so easy enough to reproduce.

Issue #169, opened 2022-08-24
ios shortcuts for metube


Yumega:
the chrome addons is nice, is there a ios shortcuts for metube?

rithask:
I can help with this

Yumega:
> I can help with this

please help😄

rithask:
There you go. Do let me know if anything goes wrong.

https://www.icloud.com/shortcuts/f1548df15b734418a77a709103bc1dd5

Yumega:
> There you go. Do let me know if anything goes wrong.
> 
> [icloud.com/shortcuts/f1548df15b734418a77a709103bc1dd5](https://www.icloud.com/shortcuts/f1548df15b734418a77a709103bc1dd5)

Thanks a lot , it works very well

rithask:
Okay. I'll create a PR to add this to README

rithask:
https://www.icloud.com/shortcuts/a1bf01b290fd46088da3621037f9e309

This should work. Can you check it?

rithask:
i've wanted a shortcut like that for a long time but i don't think that's possible since the yt-dl script requires a linux environment to run which is not possible with iOS Shortcuts. 

wil let you know if i found some solution

rithask:
@Yumega checkout [SW-DLT](https://github.com/net00-1/SW-DLT). it might fit your needs

PR #277, opened 2023-06-17
Add iOS shortcut link

rithask:


rithask:
#169

alexta69:
Thanks @rithask !

rithask:
There’s a typo in the github.com/rithask URL. I just noticed it.

alexta69:
yes. fixed

TheSpyder:
I was about to log a bug saying this wasn't working anymore but it was just because I'd copy/pasted the url from Safari and it had a trailing slash. I deleted that and it works.

Might be worth adding a note about that somewhere 🤔

PR #310, opened 2023-08-13
Fix retry button issues

JLyne:
The arguments passed to `retryDownload` by the retry button do not match what the function actually expects. This causes a few issues:

- Downloads break if a custom folder is set
- Format and custom name prefix are not applied to the retried download

This PR fixes these problems by passing the entire download object to `retryDownload`, which now populates everything correctly. The `Download` interface was also updated to include all the fields actually sent by the server.

Fixes #233, #262

PR #311, opened 2023-08-13
Support loading YTDL_OPTIONS from file

### LINKED ISSUE #210: Support loading youtubeDL options from a file
Currently metube loads the configuration of ytdl via the `YTDL_OPTIONS` env variable.
This could be nicely complemented by the possibility of loading YTDL_OPTIONS from a config file.

For example if `YTDL_OPTIONS` doesn't load as a JSon file, it could be worth checking if it's a path to a file in which case that file should be loaded instead.
An alternative is to support a `YTDL_OPTIONS_FILE` env variable which point to a configuration file. If not set then `YTDL_OPTIONS` should be used.

ArabCoders:
This PR implements a way to load yt-dlp custom options from file using the same environment variable name, i also updated the README.md to reflect the changes.

This PR closes #210

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -3,6 +3,7 @@
 
 import os
 import sys
+import traceback
 from aiohttp import web
 import socketio
 import logging
@@ -12,7 +13,9 @@ import pathlib
 from ytdl import DownloadQueueNotifier, DownloadQueue
 
 log = logging.getLogger('main')
-
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG)
+    
 class Config:
     _DEFAULTS = {
         'DOWNLOAD_DIR': '.',
@@ -38,6 +41,7 @@ class Config:
     def __init__(self):
         for k, v in self._DEFAULTS.items():
             setattr(self, k, os.environ[k] if k in os.environ else v)
+
         for k, v in self.__dict__.items():
             if v.startswith('%%'):
                 setattr(self, k, getattr(self, v[2:]))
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -12,6 +12,8 @@
 from ytdl import DownloadQueueNotifier, DownloadQueue
 
 log = logging.getLogger('main')
+if __name__ == '__main__':
```
alexta69, line 15: Why has this moved here? There's no reason to move it...

ArabCoders: Because the call to Config comes before the logging level is set as such if you use debug level it wont show up. at least from my testing.

ArabCoders: @alexta69  I can move it back but the logging in the Config class will have to be changed to INFO or removed as it's won't show up otherwise.

alexta69:
@ArabCoders I reworked the PR a bit -- I was a bit concerned about trying to interpret the same env variable in two different ways (JSON and file path) so I switched to a separate env variable for specifying the file. I hope it works well for you. Thanks!

ArabCoders:
> @ArabCoders I reworked the PR a bit -- I was a bit concerned about trying to interpret the same env variable in two different ways (JSON and file path) so I switched to a separate env variable for specifying the file. I hope it works well for you. Thanks!

Oh no worries thanks for merging the support for it. If you have a bit of time could you please take look at https://github.com/alexta69/metube/pull/310 i posted screenshot of an edge case that happens when download fail.  

Thank You.

PR #312, opened 2023-08-14
Fix memory leak when many videos are queued

### LINKED ISSUE #297: Memory leak when downloading many videos.
Hello,

I use this excellent tool to archive youtube channels, and it's works perfectly for my use case. Sadly, i think i managed to break something within the browser side, however the tool itself in the container doesn't crash.

image

it keep going up and up until chrome's crashes the tab 

image

However checking the container stats does not seems to be leaking internally, which lead me to believe it's something to do with the frontend side.

image

at the time of writing this issue, there was around 200ish video queued to be downloaded.

JLyne:
Currently angular is re-rendering the entire download queue whenever a download's state changes. 

This is inefficient, but when combined with a long download queue the frequency of websocket progress updates causes many thousands of dom nodes to be created very quickly. This causes a large increase in memory usage, which in my experience continues to worsen until the tab is closed.

This PR adds a `trackBy` function to the `ngFor` of each half of the queue to resolve this by preventing unnecessary re-rendering.

Fixes #297

### DIFF
--- a/ui/src/app/app.component.html
+++ b/ui/src/app/app.component.html
@@ -94,7 +94,7 @@
       </tr>
     </thead>
     <tbody>
-      <tr *ngFor="let download of downloads.queue | keyvalue: asIsOrder" [class.disabled]='download.value.deleting'>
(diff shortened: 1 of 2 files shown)


alexta69:
Thanks @JLyne !

PR #335, opened 2023-11-03
Expose done/queue list as JSON endpoint

ArabCoders:
Hello,

I have exposed the done/queue list as JSON endpoint which helps in aiding of writing scripts to capture live streams as sometimes the added item or cronjob will miss events. Some channels i follow after the live stream ends they make the video private thus the next time the cronjob runs, it will be presented with a private video message.

I have also fixed unrelated bug that mostly happens probably in docker if you sometimes suddenly restart docker container when the container was downloading video, the queue get corrupted. This fix simply checks if the entry is actually set instead of assume it is and crashing. Example of the error this bug fixes

```
AttributeError: 'NoneType' object has no attribute 'get'
etype = entry.get('_type') or 'video'
File "/app/app/ytdl.py", line 246, in __add_entry
return await self.__add_entry(entry, quality, format, folder, custom_name_prefix, already)
File "/app/app/ytdl.py", line 293, in add
await self.add(v.url, v.quality, v.format, v.folder, v.custom_name_prefix)
File "/app/app/ytdl.py", line 207, in __import_queue
Traceback (most recent call last):
uture: <Task finished name='Task-3' coro=<DownloadQueue.__import_queue() done, defined at /app/app/ytdl.py:205> exception=AttributeError("'NoneType' object has no attribute 'get'")>
ERROR:asyncio:Task exception was never retrieved
INFO:ytdl:adding https://www.youtube.com/shorts/sample: quality='best' format='mp4' already=None folder='sample' custom_name_prefix=''
INFO:ytdl:waiting for item to download
``` 

I have changed the url and folder to prevent copystrikes


app/main.py
```diff
@@ -132,6 +132,20 @@ async def delete(request):
     status = await (dqueue.cancel(ids) if where == 'queue' else dqueue.clear(ids))
     return web.Response(text=serializer.encode(status))
 
+@routes.get(config.URL_PREFIX + 'history')
+async def list_queue(_):
+    history = [];
+    for _ ,v in dqueue.queue.saved_items():
+        obj = vars(v)
+        obj['list_type'] = 'queue'
```
alexta69, line 140: Wouldn't it be more useful to return the two lists separately, instead of one list with a type for each member? It'll make it more compact and sensible I think... I'm talking about this sort of response:
```
{
    "queue": [
        { ...item... },
        { ...item... },
    ],
    "done": [
        { ...item... },
        { ...item... },
    ],
}
```

app/main.py
```diff
@@ -132,6 +132,20 @@ async def delete(request):
     status = await (dqueue.cancel(ids) if where == 'queue' else dqueue.clear(ids))
     return web.Response(text=serializer.encode(status))
 
+@routes.get(config.URL_PREFIX + 'history')
+async def list_queue(_):
```
alexta69, line 136: Let's rename `_` to `request` just for consistency with the other calls

app/main.py
```diff
@@ -132,6 +132,20 @@ async def delete(request):
...
+    for _ ,v in dqueue.queue.saved_items():
+        obj = vars(v)
+        obj['list_type'] = 'queue'
+        history.append(obj)
+    for _ ,v in dqueue.done.saved_items():
+        obj = vars(v)
+        obj['list_type'] = 'done'
+        history.append(obj)
+
+    return web.Response(text=json.dumps(history))
```
alexta69, line 147: If you restructure the response as suggested above, then you don't need to add a property to the object (`list_type`), and then you can call `serializer.encode(history)` here instead of `json.dumps(history)`, and if you do that, you don't need to convert the objects to dicts above with `vars()`.. you can probably just use the return value of `save_items()` directly, it'll make the code much shorter

app/main.py
ArabCoders: Indeed good idea, ills implement that and remove the `list_type` key

app/main.py
ArabCoders: Alright i'll make a patch

app/main.py
ArabCoders: Thanks for your tips, i'll make the changes. I have one question, metube does't store the error message in the queue does it? if i restart the container the list is still there, however the error become null

ArabCoders:
@alexta69 I've implemented the suggested fixes.

app/main.py
```diff
@@ -132,6 +132,17 @@ async def delete(request):
     status = await (dqueue.cancel(ids) if where == 'queue' else dqueue.clear(ids))
     return web.Response(text=serializer.encode(status))
 
+@routes.get(config.URL_PREFIX + 'history')
+async def history(request):
+    history = { 'done': [], 'queue': []}
```
alexta69, line 137: Maybe you can do it all like this --
```
history = { 'done': [v for k, v in dqueue.queue.saved_items()], 'queue': [v for k, v in dqueue.done.saved_items()]}
```
Can you see if this works? You probably don't even need to call `vars()`. But you may need to switch from `json.dumps` to `serializer.encode()` below.

ArabCoders: Done

ArabCoders:
@alexta69 i started with using the same serializer however it was complaining about DownloadInfo cannot be JSON encoded

alexta69:
> @alexta69 i started with using the same serializer however it was complaining about DownloadInfo cannot be JSON encoded

So now with the latest changes it works OK?

ArabCoders:
> > @alexta69 i started with using the same serializer however it was complaining about DownloadInfo cannot be JSON encoded
> 
> 
> 
> So now with the latest changes it works OK?

From my limited testing with sample data yeah.

alexta69:
Perfect. Thanks for your contribution!

ArabCoders:
> Perfect. Thanks for your contribution!

Thank you. Now i wish to tackle a way to be able to download multiple videos at the same time. As i download mostly live streams from youtube and usually the streamers start at or close to each other. Do you have an interest in that?

alexta69:
I would merge such a PR, sure. A configuration variable would need to be added that limits the number of concurrent downloads, and I think the safest would be to default to 1, and let users set it if needed, because I'm afraid that concurrent downloads opened by default might get people throttled (maybe not on youtube but on other platforms, who knows).

Issue #339, opened 2023-11-07
Add option for auto download.


nodew:
Add a new option along with current Format and Quality, because in a page, there might be a few videos, but user might only what to download one or two of them, then user should be able to turn off the auto download feature, after clicking on `Add`, the main page will only show all available items instead of starting download, user can choose which they what to download.

PikuZheng:
Do you mean when adding a list, all videos should be added to the queue so that users can cancel unnecessary downloads in time?

nodew:
Yes, for example:

| Name | action |
|-----------|-----------|
| video a | `download` / `cancel` / ... |
| video b | `download` / `cancel` / ... |

User can click on the `download` button to start download.

alexta69:
Fixed in #353

PR #353, opened 2023-12-09
Add option for auto starting download.

nodew:
Add a new "Auto start" option to the MeTube. 
When the setting is "No", MeTube will only extract information to the list, only when clicking on the download button, it will start download select item. 
When the setting is "Yes", there's no difference than current options.

image

It's an implementation for issue #339.


alexta69:
Thanks @nodew for your contribution!

timreuscher:
Noticing a couple of things after the implementation of this (even after the default was set to Yes):

1. The Chrome extension no longer autostarts (even with the default set to Yes). For someone not monitoring Github I simply thought Metube was broken. Maybe an option in Docker for AUTOSTART with a potential value of Always that eliminates the GUI option to revert to old behavior for those who want it would be good?
2. It seems like the only way to delete something from the Downloading queue is to start the download, then delete it.

nodew:
Hi @timreuscher I'm sorry for breaking your scenario, but I'm not aware of the use case of Chrome extension, which extension you refer to?

timreuscher:
It's called Metube Downloader - https://chromewebstore.google.com/detail/metube-downloader/fbmkmdnlhacefjljljlbhkodfmfkijdh

Very convenient. Just click a button and it sends the page you're on to Metube.

nodew:
@timreuscher thanks, I'll have a try.

malikamg:
I've the same issue. previously it was working fine. two days ago all download send to metube using chrome add-on was unable to autostart

nodew:
I made a fix, see #355

timreuscher:
Issue seems solved by the fix. Thanks!

PikuZheng:
Hi, I'd like to know is it an issue that I can not delete a task that didn't start automatically?

timreuscher:
> Hi, I'd like to know is it an issue that I can not delete a task that didn't start automatically?

I seem to remember this being the case, and the workaround was to pause the download, then delete it. Might want to start a new issue on that.

PR #361, opened 2023-12-30
PWA improvements

philipphutterer:
This PR adds support for the Web Share Target API (see https://developer.chrome.com/docs/capabilities/web-apis/web-share-target, https://developer.mozilla.org/en-US/docs/Web/Manifest/share_target). This allows you to install the Web App on Android using the Chrome Browser, and share links directly via the System share menu. The prerequisite for this is that the app is served via HTTPS (e.g., via a reverse proxy).

Also, I moved the icons to the frontend assets for clean up.


ui/src/app/app.module.ts
```diff
@@ -27,7 +28,13 @@ import { NgSelectModule } from '@ng-select/ng-select';
     NgbModule,
     HttpClientModule,
     FontAwesomeModule,
-    NgSelectModule
+    NgSelectModule,
+    ServiceWorkerModule.register('custom-service-worker.js', {
```
alexta69, line 32: How does adding a service worker affect the regular scenario of loading the application in a web browser? Is there anything different that the users are seeing (popups asking for notification permissions or anything like that)?

ui/src/custom-service-worker.js
```diff
@@ -0,0 +1,38 @@
+const URL_PATTERN =
+  /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/gi;
```
alexta69, line 2: What are we trying to match here? Can you maybe add a comment with an explanation of the structure we're expecting, just so humans don't have to parse the RE to understand this...

philipphutterer: Since the `add` endpoint is expecting a URL, and the text from the share menu often contains more than just the URL (additional text), this extracts the URL from the shared text. I know it is quite unreadable, and probably even incomplete, for URLs with other schemas than HTTP. Do you have another idea for this?

alexta69:
Thanks @philipphutterer !

PR #367, opened 2024-01-02
Pass paths also when extracting metadata

### LINKED ISSUE #363: `paths` parameter not passed to API when extracting info
After some debugging fun today I came across the following discrepancy. I'm not sure if I'm missing something and this is actually intentional, but it's causing me isuses.

There seem to be two calls two the yt-dlp API, [here](https://github.com/alexta69/metube/blob/master/app/ytdl.py#L85) when downloading videos and [here](https://github.com/alexta69/metube/blob/master/app/ytdl.py#L219) when extracting metadata. The first call passes `paths`, but the second one does not.

When downloading a Youtube playlist with `"writethumbnail": True` the second call will therefore attempt to save the **playlist** (not video) thumbnail to `./` which in my setup ([Truecharts' metube](https://github.com/truecharts/charts/tree/master/charts/stable/metube) on TrueNAS Scale) results in a permission denied OSError.

Is there a reason the second call could not look like this, also passing `paths`?

```python
    def __extract_info(self, url):
        log.info('Extracting info')
        log.info(self.config)
        return yt_dlp.YoutubeDL(params={
            'quiet': True,
            'no_color': True,
            'paths': {"home": self.config.DOWNLOAD_DIR, "temp": self.config.TEMP_DIR},
            'extract_flat': True,
            'ignore_no_formats_error': True,
            **self.config.YTDL_OPTIONS,
        }).extract_info(url, download=False)
```

I could work around this for now, by passing `paths` in YTDL_OPTIONS but it seems counter-intuitive that this is necessary, and it will also overwrite the download directory selection from the UI.

(Edit: Of course I'm happy to create the PR if it helps and there aren't any downstream effects I'm overlooking by performing this change)

JonasR:
Fixes #363 

In some cases, such as when writing thumbnails, this call can still cause creation of non-temporary files on disk which should then also be placed in DOWNLOAD_DIR

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -222,6 +222,7 @@ class DownloadQueue:
             'no_color': True,
             'extract_flat': True,
             'ignore_no_formats_error': True,
+            'paths': {"home": self.config.DOWNLOAD_DIR, "temp": self.config.TEMP_DIR},
             **self.config.YTDL_OPTIONS,
         }).extract_info(url, download=False)
 


JonasR:
I was undecided of whether to further align the two calls.

I guess including `'socket_timeout': 30` wouldn't hurt either. Similarly for `outtmpl`, but neither are directly concerned with this issue, so I left them out.

alexta69:
Perhaps if we add more parameters (to create further alignment) we should have them in a common place somewhere. But doesn't matter for now.
Thanks!

PR #370, opened 2024-01-04
Fix PWA manifest for relative paths

philipphutterer:
In my recent PR for the PWA share target I made a mistake with absolute instead of relative paths. This lead to incorrect behavior with URL paths other than root.
This PR fixes that and adds a short comment to the URL RegExp as discussed (https://github.com/alexta69/metube/pull/361#discussion_r1438685903).


alexta69:
Thanks @philipphutterer !

Issue #383, opened 2024-01-23
can't cancel download except started


PikuZheng:
> Hi, I'd like to know is it an issue that I can not delete a task that didn't start automatically?

I seem to remember this being the case, and the workaround was to pause the download, then delete it. Might want to start a new issue on that.

_Originally posted by @timreuscher in https://github.com/alexta69/metube/issues/353#issuecomment-1905852798_

PikuZheng:
Thanks @vkartk !

vkartk:
@PikuZheng  welcome :)

PR #385, opened 2024-01-26
Fix cancellation & deletion for non-started downloads (#383)

vkartk:
Currently, users cannot cancel or delete tasks if they haven't been started automatically. This PR implements proper handling for non-started downloads, allowing users to cancel and delete them before downloading begins. This also resolves the "requested cancel for non-existent download" warning.


alexta69:
Thanks again @vkartk ! :)

PR #428, opened 2024-04-28
Support basic auth in Chrome bookmarklet

apeckham:
Updated the MeTube bookmarklet to include `xhr.withCredentials=true` in the AJAX request. This allows the bookmarklet to work when MeTube is behind basic auth.

Only applied this change to the Chrome version of the bookmarklet since that is the environment I was able to test in.

Thank you for metube 📺


alexta69:
Thanks @apeckham !

PR #430, opened 2024-05-02
add `PUBLIC_HOST_URL` and `PUBLIC_HOST_AUDIO_URL` config key

### LINKED ISSUE #429: [feature] set a public url for hosting
I want my MeTube instance to be only accessible in the local network, but the output folder gets served by caddy publicly. Could you add a environment variable for a public host url? So that the links in the UI don't point to `/downloads/video.mp4` but `https://example.com/video.mp4`?

Thanks a bunch

vaaski:
resolves #429

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,8 +35,6 @@ ENV UMASK=022
 ENV DOWNLOAD_DIR /downloads
 ENV STATE_DIR /downloads/.metube
 ENV TEMP_DIR /downloads
-ENV PUBLIC_HOST_URL download/
-ENV PUBLIC_HOST_AUDIO_URL audio_download/
 VOLUME /downloads
 EXPOSE 8081
 CMD [ "./docker-entrypoint.sh" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ ENV UMASK=022
 ENV DOWNLOAD_DIR /downloads
 ENV STATE_DIR /downloads/.metube
 ENV TEMP_DIR /downloads
+ENV PUBLIC_HOST_URL download/
 VOLUME /downloads
 EXPOSE 8081
 CMD [ "./docker-entrypoint.sh" ]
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -35,6 +35,8 @@ ENV UMASK=022
 ENV DOWNLOAD_DIR /downloads
 ENV STATE_DIR /downloads/.metube
 ENV TEMP_DIR /downloads
+ENV PUBLIC_HOST_URL download/
```
alexta69, line 38: No need to set defaults here, since they're already set in `app/main.py`. Here we only override configurations that are specifically set for Docker (and are different in `app/main.py`).

app/main.py
```diff
@@ -24,6 +24,8 @@ class Config:
         'DELETE_FILE_ON_TRASHCAN': 'false',
         'STATE_DIR': '.',
         'URL_PREFIX': '',
+        'PUBLIC_HOST_URL': 'download/',
+        'PUBLIC_HOST_AUDIO_URL': 'audio_download/',
```
alexta69, line 28: If the value is changed to another path within the container, it won't work, because the routes are hard-coded in this file on line 212 and 213. There needs to be some solution for this.

ui/src/app/app.component.ts
```diff
@@ -233,9 +233,9 @@ export class AppComponent implements AfterViewInit {
   }
 
   buildDownloadLink(download: Download) {
-    let baseDir = 'download/';
+    let baseDir = this.downloads.configuration["PUBLIC_HOST_URL"] ?? 'download/';
```
alexta69, line 236: These variables already have defaults, so it seems we don't need a fallback to a default here.

app/main.py
vaaski: Not sure what L212-213 are used for but with my setup I forego those paths completely since I serve the files somewhere else (in this case a separate caddyserver)

As far as I can tell those would only apply for the default host values, but I guess a user might want to change the path prefix from `download/` to something like `output/`, is it that what you mean?

alexta69: Yes. But thinking about it again, there isn't really a need to support that. So we'll leave it as it is.

PR #431, opened 2024-05-02
Don't convert audio if necessary

jfsimoneau:
To avoid unnecessary conversion, yt-dlp can try to download the requested format first.


alexta69:
Thanks @jfsimoneau !

PR #440, opened 2024-05-21
Update MP4 format string for iOS compatibility

jgoguen:
iOS has strict requirements for video files, requiring h264 or h265 video codec and aac audio codec in MP4 container. This update to the MP4 format string tries to get a fully compatible file first, followed by the right video codec and any M4A audio (audio is much faster to convert if needed), and then falls back to the original behaviour of getting the best available MP4 video and M4A audio.


alexta69:
Can you please add a comment to the code that explains this, for future editors? Even the text above as is, just to make the logic there clear.

Issue #448, opened 2024-06-06
Download audio as .flac format


jannme310103:
Hello there,

I'm wondering if it's possible to add a option in the format Drop Down Menu that allows you to download audio as a .flac file. Is this possible?

Greetings 
Jann

PikuZheng:
yes...but can you confirm your ffmpeg support flac encode?

on my pc---->
```
localhost:~# docker run -it --rm metube:20240522v2 '/usr/local/bin/ffmpeg' '-codecs' | grep flac
 DEAI.S flac                 FLAC (Free Lossless Audio Codec)                      <--------------EA means support flac audio encode
```

jannme310103:
I don't exactly know, I'll check that if I'm back at home later

jannme310103:
So I checked it, I would say it's able to encode flac
I've checked it with this command

ffmpeg -i /home/jann/Sickmodel R Mish\ I One\ And\ only\ (Uptemp ol Edit\).flac -f null -

and this is the output:



ffmpeg version 5.1.4-0+rpt3+deb12ul Copyright (c) 2000-2023 the FFmpeg developers built with gcc 12 (Debian 12.2.0-14) configuration: --prefix=/usr --extra-version=0+rpt3+deb12ul --toolchain=hardened --incdir=/usr/include/aarch64-linux-gnu --enable-gpl --disable-stripping --disabl e-mmal --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-li bbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --en able-libdavld --enable-libflite --enable-libfontconfig --enable-libfreetype --enab le-libfribidi --enable-libglslang --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt - -enable-libopus --enable-libpulse --enable-librabbitmq --enable-librist --enable-l ibrubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspe ex --enable-libsrt --enable-libssh --enable-libsvtavl --enable-libtheora --enable-
[...]

PR #449, opened 2024-06-07
add flac format for audio

PikuZheng:
for #448


jannme310103:
I just updated my MeTube Docker container to the latest version. However, the option to download audio as a .flac file still doesn't appear?

alexta69:
If you have the latest version then you're supposed to see this:
image

jannme310103:
Now I see it, nice work guys, thank you

Issue #470, opened 2024-07-14
basic auth not asking for credentials


alegru:
I'm using MeTube behind a reverse proxy that manages TLS and basic auth for me. Authelia is overkill for what I'm doing.
Both nginx and Traefik proxy exhibit the same behaviour:
After successful login, the service worker connection will cause the browser to store cookies after a few seconds.
End the session by closing the browser. The next time you try to login, there is no prompt for the credentials.
You need to either use private mode or force delete cookies.
From what I gathered during my investigation, the problem is that the site manifest is not served correctly when using basic auth.
The angular repository had the [solution](https://github.com/angular/angular/issues/23605#issuecomment-652474956):
[developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin#example_web_manifest_with_credentials)

PR #471, opened 2024-07-14
Fix basic auth not asking for credentials

alegru:
Fixes: #470
I tested both Firefox and Chromium, with and without proxy.
After changing the manifest origin setting, on Firefox everything works correctly.
Sadly, Chromium still has problems, issues are open for years now.
In any case, the fix shouldn't have impact on users without basic auth and respects the spec.


alexta69:
Thanks @alegru !

PR #472, opened 2024-07-15
inject manifest attribute after building

alegru:
Hi, sorry to bother you again. Looks like I underestimated web development quirks...
When I tested my change by editing the compiled index.html, everything worked as described.
Testing the latest release, the index.html and manifest is cached by the browser, which again causes it to omit the basic auth challenge.
Looks like the hash-based cache-busting mechanic bit me here. The solution seems to be to introduce the manifest attribute after building.
Or changing the hash for index.html in ngsw.json after building.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN sed -i 's/\r$//g' docker-entrypoint.sh && \
 
 COPY app ./app
 COPY --from=builder /metube/dist/metube ./ui/dist/metube
+RUN sed -i 's/webmanifest"/webmanifest" crossorigin="use-credentials"/' ./ui/dist/metube/index.html
 
 ENV UID=1000
 ENV GID=1000
--- a/ui/src/index.html
+++ b/ui/src/index.html
@@ -11,7 +11,7 @@
   <meta name="msapplication-config" content="assets/icons/browserconfig.xml">
   <link rel="shortcut icon" href="favicon.ico">
   <meta name="msapplication-TileColor" content="#da532c">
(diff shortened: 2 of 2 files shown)


alexta69:
I'm not sure I understand what you're trying to do, in my copy of the new container I do have `crossorigin="use-credentials"` in my `index.html`, so why is this change needed?

alegru:
Okay, that's what I thought, too. `crossorigin="use-credentials"` is correctly set in the `index.html`.
If I clear my cookies and cache, and login to my basic auth MeTube instance, this is what the browser caches:
bad

Editing the `index.html` after compilation with sed or similar, the cache looks like this:
correct
Note the absence of `index.html` and `manifest.webmanifest` in the second image.
This is what I believe causes the browser to not ask for the index.html, which means the user is stuck on "Connecting to server" until the cache is cleared.

It is strange that by editing the file after compilation, everything works correctly. I suspect that on compilation, Angular stores a hashmap of all files in `ngsw.json` due to `"outputHashing": "all"` in `angular.json`. This hash doesn't match if the file is edited after compilation, which is why the browser always asks for a fresh copy.
Thoughts?

alexta69:
I'm not sure I understand the caching logic (I'm not familiar with ngsw.json or what is its impact). But I don't see a problem if the browser caches index.html. Why would that be an issue? The file is static, it makes sense to cache it.

alegru:
I'm not sure either, I just shared my findings that it breaks basic auth for users. Anyway, I bit the bullet and setup Authelia; I can confirm everything works as expected with the new auth mechanism. I'm gonna close this pull request since it's no longer relevant. Thank you for your awesome software!

Issue #382, opened 2024-01-23
Download whole channel


gregory-chatelier:
Given a channel URL https://www.youtube.com/@7jourssurTerre/playlists for example.

You can manually open a terminal in the container and run a command to download the whole channel with a directory structure :
```shell
yt-dlp --yes-playlist -f best -ciw -o "/downloads/youtube-dl/%(uploader)s/playlists/%(playlist)s/videos/%(playlist_index)s - %(title)s.%(ext)s" -v "https://www.youtube.com/@7jourssurTerre/playlists"
``` 
I would have liked this to happen when I gave the channel URL in the IHM

PikuZheng:
set YTDL_OPTIONS={"extract_flat": "discard_in_playlist"} maybe?
what's -ciw means?

gregory-chatelier:
I got the exact command from this thread :https://askubuntu.com/questions/1167595/use-youtube-dl-to-download-all-the-playlists-of-a-channelyt-dlp is a fork so I assumed the options are the same. The options : https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#usage-and-options The options -ciw are used with specific purposes: -c: This option, also written as --continue, allows resuming downloads that were interrupted or partially completed. If a download is stopped for any reason, using this option will resume the download from where it left off, instead of starting over. -i: This option, also written as --ignore-errors, tells YouTube-DL to continue downloading other videos in the playlist even if one video fails to download. Without this option, YouTube-DL would stop the entire process if it encounters an error with a particular video. -w: This option, also written as --no-overwrites, prevents overwriting files that already exist in the download destination. If a file with the same name already exists, YouTube-DL will not download it again.   Envoyé à partir de Courrier pour Windows De : PikuZhengEnvoyé le :mardi 23 janvier 2024 10:55À : alexta69/metubeCc : Grégory Chatelier; AuthorObjet :Re: [alexta69/metube] Download whole channel (Issue #382) set YTDL_OPTIONS={"extract_flat": "discard_in_playlist"} maybe?what's -ciw means?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

PikuZheng:
This requires always assuming that a playlist is still a playlist after parsing. which is a little complicated

neuleo:
Will this be implemented?

alexta69:
[...]

Issue #366, opened 2024-01-02
Error adding URL: 500 (KeyError: 'id') - But Video can be downloaded when using yt-dlp directly


Hely0n:
I just tested MeTube a little bit and I recognized an error for a non-youtube Video that normally works with yt-dlp.

When I try to add it, a popup appears, saying "Error adding URL: 500 Internal Server Error
Server got itself in trouble" 

And the logs say:

> INFO:ytdl:playlist detected with 1 entries
> ERROR:aiohttp.server:Error handling request
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
>     resp = await request_handler(request)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_app.py", line 543, in _handle
>     resp = await handler(request)
>            ^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/app/main.py", line 125, in add
>     status = await dqueue.add(url, quality, format, folder, custom_name_prefix, auto_start)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/app/ytdl.py", line 313, in add
>     return await self.__add_entry(entry, quality, format, folder, custom_name_prefix, auto_start, already)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/app/ytdl.py", line 270, in __add_entry
>     etr["playlist"] = entry["id"]
>                       ~~~~~^^^^^^
> KeyError: 'id'

The link I tried is only available in Germany, so I think it won't help for debugging, but maybe the logs will be enough.

PikuZheng:
Yes, non-YouTube websites may not be able to read the video ID. Have you tried a single download instead of a playlist?

Hely0n:
Ah, that must be the culprit. The video on this page is internally split up into multiple parts, where the video player may insert ads between them. Thats why one have to use the "--concat-playlist always" flag of yt-dlp.

PR #496, opened 2024-08-18
Playlist strict mode, limits and fixes

labmonkey:
Added:
Two new options that make it safer to download playlists
- Strict Playlist mode. I tend to download videos that are part of playlists (such as `https://www.youtube.com/watch?v=VIDEO_ID&list=PLAYLIST_ID`). Unfortunately default behaviour of ytdlp in such case is to treat the link as a url and download whole playlist, even that my intention was to download just a single video.
  With the Strict Playlist mode that can be enabled/disabled by default by env variable it will now force such urls to be treated as single video url and download playlists only if directly requested (url points to `https://www.youtube.com/playlist?list=PLAYLIST_ID`). This saves user from having to manually remove the PLAYLIST_ID part from url if he wanted to achieve that. Can be also changed in the UI.
- Playlist item limit. Simply limits the amount of playlist items that can be downloaded at once. This prevents unintentional download of huge playlists or simply allows to paste playlist url and download only recent items. Can be also changed in the UI.
- `OUTPUT_TEMPLATE_PLAYLIST` for playlist specific template, mainly to download playlist entries into their own `playlist_title` subdirectory 

Fixed:
- when downloading an URL such as `https://www.youtube.com/watch?v=VIDEO_ID&list=PLAYLIST_ID` even that ytdlp was downloading a whole playlist then in the UI you could only see a single job downloading multiple videos with title same as the playlist title. It was because the `etype = 'url'` check was performed last. Now those kind of urls will be processed same as a direct playlist link and each video will be added in separate job.


PikuZheng:
could this fix #382 ?

labmonkey:
> could this fix #366 or #382 ?

I thinks those are different. I would not expect these changes to have an impact.

alexta69:
Thanks @labmonkey !

alexta69:
@labmonkey can you please look at #497 ? looks like the PR introduced a regression

MattBlackOnly:
I think this should be reverted as it doesn't add anything of value and clutters the UI. 

The issues it's trying to fix are user error. 

For example, if you only want one video from a playlist, get that specific video link. 

And if you don't want to download all videos from a playlist, you can just create a new playlist with the items you do want and download that.

@alexta69

MattBlackOnly:
And if you want to download playlists into a specific folder, you just enable DONWLOAD_DIRS_INDEXABLE and create the new folder when downloading.

PR #498, opened 2024-08-21
Added batch script to automate setup in windows.

Thepathakarpit:
Just run setup.bat and with one command you can run it on your windows os.
It will make users be able to use it easily.

### DIFF
--- a/setup.bat
+++ b/setup.bat
@@ -1,0 +1,45 @@
+@echo off
+
+rem Check if Python is installed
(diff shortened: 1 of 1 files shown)


alexta69:
We don't have similar scripts for other platforms. And there's very little that's MeTube-specific here. Other users might use other methods to install pip for instance. I think we don't need to add it to this repository. Sorry and thanks

PR #500, opened 2024-08-22
removed version line from Docker Compose YAML section

volumedata21:
Removed line `version: "3"` from Docker Compose YAML example. Specifying version is now obsolete and will throw a warning message when launching docker compose file.


alexta69:
Thanks @volumedata21 !

PR #501, opened 2024-08-22
Remove unnecessary settings for playlists (to maintain Metube's Simplicity and User-Friendliness)

MattBlackOnly:
The beauty of Metube lies in its simplicity, quick setup, and user-friendly interface without excessive settings. Introducing more settings for playlists diminishes this simplicity and offers no real benefit to the majority of users.

This PR reverts the changes made by #496 and #499, based on the following reasoning:
> I think this should be reverted as it doesn't add anything of value and clutters the UI.
> The issues it's trying to fix are user error.
> 
> For example, if you only want one video from a playlist, get that specific video link.
> And if you don't want to download all videos from a playlist, you can just create a new playlist with the items you do want and download that.
> 
> And if you want to download playlists into a specific folder, you just enable CUSTOM_DIRS and create the new folder when downloading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ docker run -d -p 8081:8081 -v /path/to/downloads:/downloads ghcr.io/alexta69/met
 ## Run using docker-compose
 
 ```yaml
-version: "3"
 services:
   metube:
     image: ghcr.io/alexta69/metube
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ docker run -d -p 8081:8081 -v /path/to/downloads:/downloads ghcr.io/alexta69/met
 ## Run using docker-compose
 
 ```yaml
+version: "3"
 services:
   metube:
     image: ghcr.io/alexta69/metube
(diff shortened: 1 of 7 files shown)


alexta69:
@MattBlackOnly , we're not opening a PR war here. If #496 was approved, then it was approved.

MattBlackOnly:
@alexta69 It’s your project, so you can decide what changes to make. 

But let’s be clear: unnecessary feature bloat benefits no one. Features should only be added if they genuinely fit the general use case and benefit everyone. 

The points I raised are entirely valid. Had they been reviewed properly, you’d find it hard to disagree.

alexta69:
Very few features benefit literally everyone, because different people have different use cases, and use different subsets of features. Including only features that benefit everyone will make sure that the software remains incomplete for the majority of use cases.

I disagree that a video link that also has its playlist source indicated in the URL should download the playlist. Video links should download videos, and playlist links should download playlists. So I think it us a very valid fix. I would even set it as true by default, frankly, maybe I'll push this change.

And I disagree that sending people to create their playlists is a sensible workaround, it's cumbersome (and may be even impossible on some platforms, remember that this is not only for YouTube), so I welcome that addition as well.

Regarding OUTPUT_TEMPLATE_PLAYLIST, we already have output settings for other things (such as different directory for audio, or different template for chapters), so this also makes sense to me.

alexta69:
You don't need to go out of your way, you just need to copy the URL from the address bar in the browser, which is the first thing many people think of doing.

I agree that "strict playlist mode" is not a descriptive name. I think I'll remove this configuration altogether and just have this mode as the only one.

Thank you for your input.

PR #505, opened 2024-08-29
 Add the ability to be configured as an HTTPS service

VergilGao:
- Add a supports_reuse_port method to check if the operating system supports reuse port. 
    This fixes the runtime error when using source code on Windows systems.
- Add the ability to be configured as an HTTPS service.
    This is very useful for users who want to communicate over HTTPS between the reverse proxy server and Metube, as it avoids potential man-in-the-middle attacks (although this may not appear to be a threat).


VergilGao:
This is very useful for users who want to communicate over HTTPS between the reverse proxy server and Metube, as it avoids potential man-in-the-middle attacks (although this may not appear to be a threat).

VergilGao:
Hi @alexta69 
I created this PR to add a new feature, as indicated in the title, to enable HTTPS access for the MeTube server.
I am running an Nginx server on another host within the local network, which is responsible for reverse proxying all my internal services.
If HTTPS communication can be used between Nginx and MeTube, it will obviously protect them from man-in-the-middle attacks (although the possibility is minimal).
Even so, I still believe that using HTTPS instead of HTTP is a good habit. Moreover, this PR only introduces an option and does not mandate users to use HTTPS.
If you have any questions, we can continue to communicate.

alexta69:
@VergilGao Thanks for the PR, I'm on vacation at the moment and that's why I'm not handling it. I get back later this week and will look at it.

alexta69:
Thanks @VergilGao !

PR #520, opened 2024-10-02
fix: provide default value for title when absent

phoenix727:
Issue: When trying to download any series from tubitv, there is an error that appears and I'm not able to download the series. I am able to download individual videos from within the playlist.

Reason: When creating the DownloadInfo object, there is an error thrown due to the title not existing on the entry. Adding a fallback fixes this error and allows things to proceed as expected.


alexta69:
Thanks @phoenix727 !

Issue #478, opened 2024-07-26
I cna't access metube using https (with message "Connetcing to server...") 


kilo20:
I installed metube using docker. When I run metube in local network there is no problem. But when I access metube using https, metube don't run, showing message "conneting to server...". I installed authelia to authencicate the metube.

The issue uccurs when I access metube using authelia. But after I empt the cache on browser(regardless safari, chrome, firefox and so on), I can access the metube normally. I use nginx proxy manager.

And using Pirvate Browsing(in safari), Sectet Window mode(in google Chrome) etc, I can access the metube normally. So I suppose that the problem is associated with the cookie. How can I solve the issue?

<img width="1150" alt="screen" src="https://github.com/user-attachments/assets/1a76fdea-cd4e-4a34-998d-988afeb3bf8d">

PikuZheng:
config file of nginx please

kilo20:
> config file of nginx please

I didn't install nginx. I only installed nginx-proxy-manager using docker. In other words, my https access environment is nginx-proxy-manger + cloudflare DNS + austelia.

PikuZheng:
I see. I believe it's related to austelia because I can reproduce this issue. The root cause may be that austelia does not correctly handle the identity authentication of the WSS protocol.
I'm not familiar with this, sorry I can't be of more help :(

alexta69:
Fixed in #548

Issue #459, opened 2024-06-21
WebSocket fails when running behind Authelia & Caddy


danshat:
Apparently after (relatively) recent updates Metube fails to work correctly while running behind Authelia and a reverse proxy.

This is how the page looks, showing "Connection to server" message.

![](https://github.com/alexta69/metube/assets/13084473/e3a3eeb7-7f9b-42b8-95c3-cb6e824c425c)

Network tab shows the following. 

![](https://github.com/alexta69/metube/assets/13084473/4002e059-a9eb-4469-9bbc-0e416a1b4fab)

docker-compose.yml:

```
 metube:
        container_name: metube
        image: alexta69/metube:latest
        volumes:
            - *******:/downloads
        restart: unless-stopped

```

Authelia logs show this repeatedly:

```
time="2024-06-21T00:07:51Z" level=info msg="Access to https://metube.example.com/socket.io/?EIO=4&transport=polling&t=P0u7k_t (method GET) is not authorized to user <anonymous>, responding with status code 302 with location redirect to https://auth.example.com/?rd=https%3A%2F%2Fmetube.example.com%2Fsocket.io%2F%3FEIO%3D4%26transport%3Dpolling%26t%3DP0u7k_t&rm=GET" method=GET path=/api/authz/forward-auth remote_ip=192.168.0.1
```

Caddyfile configuration:

```
metube.example.com {
        forward_auth authelia:9091 {
                uri /api/authz/forward-auth
                copy_headers Remote-User Remote-Groups Remote-Name Remote-Email
        }
        reverse_proxy metube:8081
}
```

I understand this is likely an issue with misconfigured Caddy instance. However if someone could possibly help me with this it would be great.

PikuZheng:
can you try it without https to see if it's ssl's issue?

PikuZheng:
After some google, I think this is a problem with Caddy proxying the wss protocol. Maybe changing the Authelia authentication method from http headers to cookies will help. Or open the wss protocol in Authelia (no authentication required)

danshat:
> can you try it without https to see if it's ssl's issue?

Authelia does not allow HTTP traffic: `Target URL 'http://metube.example.com/' has an insecure scheme 'http', only the 'https' and 'wss' schemes are supported so session cookies can be transmitted securely`.

> After some google, I think this is a problem with Caddy proxying the wss protocol.

Caddy's official documentation states:
```
example.com {
	@websockets {
		header Connection *Upgrade*
[...]

PR #548, opened 2024-12-08
Exclude index.html from service worker

mazzz1y:
The Service Worker caches the index.html file. When MeTube is behind an auth proxy, the browser opens index.html from the cache instead of returning a 401 status(in the case of basic auth) or redirecting for OAuth2 forward proxies(such as Authelia, Authentik, OAuth2-Proxy, etc).

This PR excludes index.html from the Service Worker, ensuring that the first request is always fresh and returns the correct status code.

This addresses the following issues:
- https://github.com/alexta69/metube/issues/478
- https://github.com/alexta69/metube/issues/459

### DIFF
--- a/ui/ngsw-config.json
+++ b/ui/ngsw-config.json
@@ -8,7 +8,6 @@
       "resources": {
         "files": [
           "/favicon.ico",
-          "/index.html",
           "/manifest.webmanifest",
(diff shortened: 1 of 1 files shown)


alexta69:
Thanks @mazzz1y

PR #553, opened 2024-12-24
Non-docker friendly README.md

anonhostpi:
Docker has a high RAM overhead and isn't ideal for lightweight webservers or servers running with limited or cost-based resources like AWS EC2

Added more details catered to non-docker environments
- Added note about pyenv
- Added missing git clone statement
- Added options for controlling port and listening address from the command line in place of a reverse-proxy or DNAT

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,8 +261,10 @@ MeTube development relies on code contributions by the community. The program as
 ## Building and running locally
 
 Make sure you have node.js and Python 3.11 installed.
+- On systems with system-python packages (like Ubuntu), it is recommended that you use pyenv to manage python: https://github.com/pyenv/pyenv
 
 ```bash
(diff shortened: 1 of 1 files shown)


anonhostpi:
This PR is doc-only and makes no changes to the codebase

alexta69:
I'm not sure this is necessary. There's no specific need for pyenv (I don't use it), and the HOST and PORT values are the default anyway so this is redundant.

anonhostpi:
...

Doesn't work out of the box with Ubuntu 22's system python packages, hence pyenv.

Also you don't have _any_ docs on HOST and PORT, so they weren't redundant, you just never mentioned them. I had to scour your source code for those 2 environment variables.

I added these docs, because I spent 4 hours 2 days ago troubleshooting a 6 line installation script...

Issue #559, opened 2024-12-27
ERROR 500 - when using impersonate yt-dlp works directly but not via metube


admapop:
YTDL_OPTIONS={"cookiefile":"/cookies/cookies.txt","verbose":"true","impersonate":"chrome"}

If I enter the docker container and execute yt-dlp directly with the cookiefile and impersonate, the download works.

When I try via the UI I get

```
INFO:ytdl:adding https://www.udemy.com/course/complete-react-developer-zero-to-mastery/learn/lecture/15090998#overview: quality='best' format='any' already=None folder=None custom_name_prefix='' playlist_strict_mode=False playlist_item_limit=0
[debug] Encodings: locale UTF-8, fs utf-8, pref UTF-8, out utf-8 (No ANSI), error utf-8 (No ANSI), screen utf-8 (No ANSI)
[debug] yt-dlp version stable@2024.12.23 from yt-dlp/yt-dlp [65cf46cdd] (pip) API
[debug] params: {'quiet': True, 'no_color': True, 'extract_flat': True, 'ignore_no_formats_error': True, 'noplaylist': False, 'paths': {'home': '/downloads', 'temp': '/downloads'}, 'cookiefile': '/cookies/cookies.txt', 'verbose': 'true', 'impersonate': 'chrome', 'color': 'no_color', 'compat_opts': set(), 'http_headers': {'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8', 'Accept-Language': 'en-us,en;q=0.5', 'Sec-Fetch-Mode': 'navigate'}}
[debug] Python 3.11.11 (CPython aarch64 64bit) - Linux-6.8.0-1013-oracle-aarch64-with (OpenSSL 3.3.2 3 Sep 2024)
[debug] exe versions: ffmpeg 6.1.2 (setts), ffprobe 6.1.2
[debug] Optional libraries: Cryptodome-3.21.0, brotli-1.1.0, certifi-2024.08.30, curl_cffi-0.7.1, mutagen-1.47.0, requests-2.32.3, sqlite3-3.47.1, urllib3-2.2.3, websockets-13.1
[debug] Proxy map: {}
[debug] Request Handlers: urllib, requests, websockets, curl_cffi
[...]

PR #560, opened 2024-12-28
fix: impersonate property needs to be of the right type and not as it is being passed now via JSON

admapop:
https://github.com/alexta69/metube/issues/559

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -7,6 +7,9 @@ import asyncio
 import multiprocessing
 import logging
 import re
+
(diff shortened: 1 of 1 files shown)


app/ytdl.py
```diff
@@ -7,6 +7,9 @@
 import multiprocessing
 import logging
 import re
+
+import yt_dlp.networking
```
PikuZheng, line 11: imported but not used?

admapop: Probably auto imported by mistake

PR #570, opened 2025-01-12
Add tini as entrypoint in Dockerfile

l0f3n:
I noticed that the docker container was slow to shut down, just a little over 10 seconds. This is a strong indication that the sigint signal is not getting forwarded to subprocesses properly, and the container only shuts down once it receives sigkill. This is easily fixed by using `tini` as the entrypoint, which will, among other things, handle the signals properly. 

Here is an example before and after adding `tini`:
250112_12h22m07s_screenshot

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ COPY Pipfile* docker-entrypoint.sh ./
 # Install dependencies
 RUN sed -i 's/\r$//g' docker-entrypoint.sh && \
     chmod +x docker-entrypoint.sh && \
-    apk add --update ffmpeg aria2 coreutils shadow su-exec curl && \
+    apk add --update ffmpeg aria2 coreutils shadow su-exec curl tini && \
     apk add --update --virtual .build-deps gcc g++ musl-dev && \
     pip install --no-cache-dir pipenv && \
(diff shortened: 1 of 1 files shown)


alexta69:
How does this work -- tini gets the signal, and forwards it to the process, if the process doesn't handle it (and keeps running) then aren't we left in the same situation? why is it shutting down when it gets the signal from tini, and doesn't if it gets the signal from docker?

l0f3n:
I know nothing more than what I just read here: https://github.com/krallin/tini/issues/8

But the gist of it is these few quotes (he is talking about Jenkins as an example):

> Second, if Jenkins runs as PID 1, then it may not receive the signals you send it! That's a subtlety in PID 1. Unlike other unlike processes, PID 1 does not have default signal handlers, which means that if Jenkins hasn't explicitly installed a signal handler for SIGTERM, then that signal is going to be discarded when it's sent (whereas the default behavior would have been to terminate the process).

> As to whether you should be using Tini. [...] Is whatever process you exec in your entrypoint registering signal handlers? A good way to figure this out might be to check whether your process responds properly to e.g. docker stop (or if it waits for 10 seconds before exiting)

So I guess another way to fix this is to add explicit signal handlers, but why bother?

PR #584, opened 2025-02-02
display yt-dlp version in webui

PikuZheng:
In order to clearly know the current yt-dlp version in use and to distinguish whether the issue is known to yt-dlp, the current yt-dlp version number is displayed at the bottom of the web page.

image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -12,6 +12,7 @@ import json
 import pathlib
 
 from ytdl import DownloadQueueNotifier, DownloadQueue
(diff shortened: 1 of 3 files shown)


TD-Community:
Like this one. Also why not display the MeTube version itself. I always wonder what version I have running and which versions are released.

alexta69:
Thanks @PikuZheng , I totally forgot about this PR until now, sorry it took so long :)

PR #588, opened 2025-02-07
Update navbar with new MeTube logo

mynameismaurizio:
Made and added logo SVG to assets/icons
Updated navbar with SVG logo replacing text
Maintained clickable functionality of navbar brand


alexta69:
There are already logo icons there, why do we need this one? Also, the logo includes the name "MeTube" as text, I don't see a reason for keeping text in the logo. And it uses a font called "YouTube Sans Dark" which I'm not sure where it should come from.
Bottom line, we could add the logo before the "MeTube" text in the navbar, but we should do it using the existing icons, and keep the text as is.

PR #602, opened 2025-03-02
Added different download options and url import/exports

rdiaz738:
Added the option to configure downloads as either sequential, limited or concurrent. Also added import, export, or copy buttons so you can add a list of urls if you have them. 
**NOTE!!!!** There is a bug where if you select all the items that are downloading and queued, and click cancel selected the next queued item will still download. I think it's because as a downloaded item is canceled the status is being changed for the next queued item and it ends up being skipped. If that's solved then it should be good

**Sequential:**
2025-03-01 18_05_38-Window

**Concurrent**
2025-03-01 18_04_21-Window

**Limited**
2025-03-01 18_02_44-Window

**Batch Imports**
2025-03-01 18_02_25-Window
2025-03-01 17_48_09-Window

**Newly added Buttons**
2025-03-01 17_47_53-Window


README.md
```diff
@@ -61,8 +61,12 @@ Certain values can be set via environment variables, using the `-e` parameter on
 * __YTDL_OPTIONS__: Additional options to pass to youtube-dl, in JSON format. [See available options here](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/YoutubeDL.py#L183). They roughly correspond to command-line options, though some do not have exact equivalents here, for example `--recode-video` has to be specified via `postprocessors`. Also note that dashes are replaced with underscores.
 * __YTDL_OPTIONS_FILE__: A path to a JSON file that will be loaded and used for populating `YTDL_OPTIONS` above. Please note that if both `YTDL_OPTIONS_FILE` and `YTDL_OPTIONS` are specified, the options in `YTDL_OPTIONS` take precedence.
 * __ROBOTS_TXT__: A path to a `robots.txt` file mounted in the container
-
-The following example value for `YTDL_OPTIONS` embeds English subtitles and chapter markers (for videos that have them), and also changes the permissions on the downloaded video and sets the file modification timestamp to the date of when it was downloaded:
```
alexta69, line 65: Was this paragraph dropped by mistake?

README.md
```diff
@@ -61,8 +61,12 @@ Certain values can be set via environment variables, using the `-e` parameter on
 * __YTDL_OPTIONS__: Additional options to pass to youtube-dl, in JSON format. [See available options here](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/YoutubeDL.py#L183). They roughly correspond to command-line options, though some do not have exact equivalents here, for example `--recode-video` has to be specified via `postprocessors`. Also note that dashes are replaced with underscores.
 * __YTDL_OPTIONS_FILE__: A path to a JSON file that will be loaded and used for populating `YTDL_OPTIONS` above. Please note that if both `YTDL_OPTIONS_FILE` and `YTDL_OPTIONS` are specified, the options in `YTDL_OPTIONS` take precedence.
 * __ROBOTS_TXT__: A path to a `robots.txt` file mounted in the container
-
-The following example value for `YTDL_OPTIONS` embeds English subtitles and chapter markers (for videos that have them), and also changes the permissions on the downloaded video and sets the file modification timestamp to the date of when it was downloaded:
+* __DOWNLOAD_MODE__ :This flag controls how downloads are scheduled and executed. There are three possible values:
```
alexta69, line 64: Can you please document what the defaults are, for this flag and the next? You can use the same style as is used for the other configuration options: Defaults to <value>.

app/main.py
```diff
@@ -152,32 +162,37 @@ async def delete(request):
...
 async def history(request):
     history = { 'done': [], 'queue': [], 'pending': []}
 
-    for _ ,v in dqueue.queue.saved_items():
+    for _, v in dqueue.queue.saved_items():
         history['queue'].append(v)
-    for _ ,v in dqueue.done.saved_items():
+    for _, v in dqueue.done.saved_items():
         history['done'].append(v)
     for _ ,v in dqueue.pending.saved_items():
```
alexta69, line 187: I see that the same whitespace mishap exists here as above, perhaps worth fixing too while you're at it

app/ytdl.py
```diff
@@ -305,20 +370,20 @@ async def __add_entry(self, entry, quality, format, folder, custom_name_prefix,
...
-
                 if auto_start is True:
-                    self.queue.put(Download(dldirectory, self.config.TEMP_DIR, output, output_chapter, quality, format, ytdl_options, dl))
-                    self.event.set()
+                    download = Download(dldirectory, self.config.TEMP_DIR, output, output_chapter, quality, format, ytdl_options, dl)
+                    self.queue.put(download)
+                    if self.config.DOWNLOAD_MODE == 'sequential':
+                        asyncio.create_task(self.__start_download(download))
+                    else:
+                        asyncio.create_task(self.__start_download(download))
```
alexta69, line 386: Looks like both the `if` and the `else` run the same code?

alexta69:
Thanks @rdiaz738 ! It's been a long-requested feature, it's great that you've implemented it.
Maybe it's worth trying to arrange the new buttons somehow into the grid that already exists under the URL bar? Right now it looks like they're detached in their own space and there's some waste of screen real estate. Maybe there needs to be a "batch" drop bar or something like that.

rdiaz738:
@alexta69 I fixed all the issues you pointed out and updated the batch button to be next to the add button. 

2025-03-04 16_15_36-Window
2025-03-04 16_15_23-Window

alexta69:
@rdiaz738 , this morning #605 was opened after this change, I think perhaps we need to give more space to the buttons (and less to the drop-downs), can you check it?

rdiaz738:
> @rdiaz738 , this morning #605 was opened after this change, I think perhaps we need to give more space to the buttons (and less to the drop-downs), can you check it?
Ya I'm on it, also working on a fix for #607 as well

Skyslycer:
Any reason why this doesn't store all URLs immediately on the server and instead slowly queues the URLs? I'm currently downloading roughly 300 videos at once (yeah I know, a bit much), but I thought I could just click "Import URLs" and it would ingest all URLs at once and store them server side, so I don't have to wait for queuing and can close the browser.

rdiaz738:
> Any reason why this doesn't store all URLs immediately on the server and instead slowly queues the URLs? I'm currently downloading roughly 300 videos at once (yeah I know, a bit much), but I thought I could just click "Import URLs" and it would ingest all URLs at once and store them server side, so I don't have to wait for queuing and can close the browser.

Pull request is up for this https://github.com/alexta69/metube/pull/963

Issue #556, opened 2024-12-26
Queue doesn't restore when container restarted


mtrolley:
I am a new user and think I have everything set up properly, but if I have queued items and restart the container they do not get restored. Here's the log when the container restarts:

```
Setting umask to 022
Creating download directory (/downloads), state directory (/config), and temp dir (/tempdownloads)
Changing ownership of download and state directories to 1000:1000
Running MeTube as user 1000:1000
INFO:main:Listening on 0.0.0.0:8081
DEBUG:asyncio:Using selector: EpollSelector
INFO:ytdl:waiting for item to download
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-3' coro=<DownloadQueue.__import_queue() done, defined at /app/app/ytdl.py:213> exception=AttributeError("'DownloadInfo' object has no attribute 'playlist_strict_mode'")>
Traceback (most recent call last):
  File "/app/app/ytdl.py", line 215, in __import_queue
    await self.add(v.url, v.quality, v.format, v.folder, v.custom_name_prefix, v.playlist_strict_mode, v.playlist_item_limit)
                                                                               ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DownloadInfo' object has no attribute 'playlist_strict_mode'
```

The host directory mounted to `/config` contains three files: `completed`, `pending`, and `queue`. The `queue` file modification time is being updated when I add a new URL, so it looks like it's being tracked properly, but not being restored properly.

PikuZheng:
same issue, but the reason is that the queue file formats of the new and old versions are different.
so, I gave up on the queue

PR #604, opened 2025-03-05
Fix: Restore queue when container restarted from old queue file

eoredthar:
In case the user tries to restart the container with an updated container but an old queue file, the system checks if the object contains the attributes `playlist_strict_mode` and `playlist_item_limit`. If they are not present, it sets them to `False` and `0`, respectively.

Fixes: https://github.com/alexta69/metube/issues/556


alexta69:
Thanks @eoredthar !

PR #636, opened 2025-04-12
Redirect to / when basePath is empty

CedricFinance:
When the path is `/share-target`, `basePath` ends up empty and the `redirect("", 303)` redirects to the JS of the service worker. I got this behaviour on Chrome 135 on MacOS.

image


alexta69:
Thanks @CedricFinance !

Issue #466, opened 2024-07-09
[Feature request] Add support for curl-cffi


janusn:
I have encountered some videos on vimeo require [curl-cffi](https://github.com/yifeikong/curl_cffi) in order to download. Is there a way to include it or an option to bundle it with the image?

PikuZheng:
The analysis of video sites is done by yt-dlp. I don't think yt-dlp will call it after adding some wheels.

janusn:
There is a [specific Impersonation section](https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#impersonation) on the readme of [yt-dlp](https://github.com/yt-dlp/yt-dlp) that mentioned how to integrate [curl_cffi](https://github.com/yifeikong/curl_cffi) as the recommended solution.

Could you please take a look?

PikuZheng:
Since it is difficult to determine the architecture, I think the easiest way is to inspect the running metube container and install it.

take x86_64 for example:
```
wget -qO- https://github.com/yifeikong/curl-impersonate/releases/download/v0.7.0/libcurl-impersonate-v0.7.0.x86_64-linux-musl.tar.gz |tar xvz -C /lib
apk add build-base gcc g++ musl-dev
pip install --no-cache-dir curl_cffi==0.5.10
```

and  determine if the installation was successful
```
yt-dlp --list-impersonate-targets
```

janusn:
Thanks for your prompt reply.

Yeah, I am ok to add a little script to install it but it has to be done every time metube is updated.

Is it possible for the container to run scripts in a specific directory every time it starts up? Like what linuxserver.io [does](https://docs.linuxserver.io/general/container-customization/)?

PikuZheng:
build with [this Dockerfile](https://github.com/alexta69/metube/blob/b0320379fe586011aa6e8f62c5383b24d1152056/Dockerfile) if you really want :)

alexta69:
Fixed in #663

janusn:
@alexta69 
As the feature #663 has been implemented, I have tried to build a metube image from the [Dockerfile](https://github.com/alexta69/metube/blob/b0320379fe586011aa6e8f62c5383b24d1152056/Dockerfile) you suggested. But there are errors that stops it from complete. Copied from dockage:
```
 ✔ Container metube  Started                                                                        1.0s 
[+] Pulling 1/0
 ✔ metube Skipped - No image to be pulled                                                           0.0s 
[+] Running 1/0
[...]

PR #667, opened 2025-05-29
Install mutagen

jonfairbanks:
Includes `mutagen` as a dependency in the Pipfile. This should resolve errors experienced when trying to embed thumbnails.


PikuZheng:
could you alos add curl-cffi please https://github.com/alexta69/metube/issues/466#issuecomment-2920454197

PR #702, opened 2025-06-28
add LOGLEVEL and ENABLE_ACCESSLOG env variables

### LINKED ISSUE #690: Change the log level?
Wonderful software!

Can there be a way to set the log level via environment variable? An INFO log message every time it updates the UI is a whole buttload of log noise. It'd be nice if I could set it to WARNING, so I can find problems in the noise of stuff :)

I tried setting an env var `LOG_LEVEL=WARNING` but that didn't seem to do anything.

Thanks again!

nixielectra:
as per title, not sure if needed or not but for someone that doesn't want too much noise in the log.

maybe can fix #690


alexta69:
I changed the default log levels, so that we don't get so many logs
Thanks for your contribution!

PR #719, opened 2025-07-17
add ytdlp-options setting in web

xerdream:

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -92,6 +92,20 @@ class Config:
                 sys.exit(1)
             self.YTDL_OPTIONS.update(opts)
 
+    def update_ytdl_options_file(self):
+        if not self.YTDL_OPTIONS_FILE:
+            log.info('YTDL_OPTIONS_FILE not set, writing to default location')
+            self.YTDL_OPTIONS_FILE = os.path.join(self.DOWNLOAD_DIR,'ytdl_options.json')
+        log.info('Writing YTDL_OPTIONS to %s', self.YTDL_OPTIONS_FILE)
+        try:
+            with open(config.YTDL_OPTIONS_FILE, 'w') as f:
+                f.write(serializer.encode(config.YTDL_OPTIONS))
+        except:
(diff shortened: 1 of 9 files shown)


alexta69:
It seems like if `YTDL_OPTIONS_FILE` wasn't set in the environment, then after the options are edited in the UI and saved to the default filename, they won't be loaded next time the container starts. I don't suppose this is the desired behavior.

At the moment the default for `YTDL_OPTIONS_FILE` is empty, meaning that a file isn't used. That's what's documented in the README and that's how the code looks and behaves. Defaults are defined at the top. We mustn't bury a default filename deep in the code.

As I see it, there are two ways out of this: either decide that if `YTDL_OPTIONS_FILE` isn't defined, the editing option is not available; or set a default for this variable (sort of like you did, except do it at the top of the file and document it in the README). But then we must allow MeTube to load without errors even if this file isn't found. And then, if the user made a mistake in the filename when configuring this variable, it won't be loaded, there won't be errors, and the user won't know why their setting isn't applying, which is not good UX.

Frankly I think both options are not ideal... Unless there's another option that makes sense, I wouldn't include such editing functionality in MeTube at all. Software that uses configuration files doesn't necessary include a text editor with the option to edit these files inside the software itself. People can just edit the configuration with their favorite editor outside of MeTube, like they do with other software...

xerdream:
Thanks for the reply, I understand

xerdream:
@alexta69 Or is it possible to add 2 functions to upload files? I don't really want the ability to edit the configuration, I want to be able to change the YTDL_OPTIONS without restarting the service or Docker

xerdream:
@alexta69  Or is it possible to add the ability to upload configuration files?  I'd actually like to be able to change the YTDL_OPTIONS without restarting the service or Docker.

alexta69:
I suppose it's possible to add code that watches the file for changes and loads it if it's been updated. I think it's a good idea.

PR #730, opened 2025-07-24
feat(deps): Enable curl-cffi for yt-dlp and update dependencies

outlook84:
This enables the curl-cffi extra for yt-dlp.

This also includes updates to other Python dependencies as a result of re-locking the Pipfile.

https://github.com/alexta69/metube/pull/667

This previous pr doesn't work when curl-cffi is needed, yt-dlp without curl-cffi extra doesn't use curl-cffi to make requests.

### DIFF
--- a/.github/workflows/update-yt-dlp.yml
+++ b/.github/workflows/update-yt-dlp.yml
@@ -25,7 +25,7 @@ jobs:
           pipenv sync
           VER=`pipenv run pip list -o | awk '$1 == "yt-dlp" {print $3}'`
           if [ -n "$VER" ]; then
-            pipenv update yt-dlp
+            pipenv update yt-dlp[default,curl-cffi]
             git config --global user.email "updater@metube"
             git config --global user.name "AutoUpdater"
             git commit -m 'upgraded yt-dlp' Pipfile.lock


PikuZheng:
Why not use the system package curl-cffi instead of the Python package? This might result in additional packages to download or compile?
On the other hand, yt-dlp only supports some specific version of curl-cffi (not sure if it has been improved now

outlook84:
> Why not use the system package curl-cffi instead of the Python package? This might result in additional packages to download or compile? On the other hand, yt-dlp only supports some specific version of curl-cffi (not sure if it has been improved now

I can't find any info of system‑packaged curl-cffi.

PikuZheng:
well...I don’t care except additional packages downloaded or compile.

I’m using pre-build package  https://github.com/PikuZheng/metube/blob/master/Dockerfile#L22-L24

alexta69:
I tried to rebase your branch on master, but I don't have permission to push to it, so you'll have to do it. Please rebase on master and then I'll merge the PR

outlook84:
> I tried to rebase your branch on master, but I don't have permission to push to it, so you'll have to do it. Please rebase on master and then I'll merge the PR

Done.

Issue #487, opened 2024-08-06
[Feature request] Tag releases using semantic versioning


namelivia:
Hello! First of all, thanks for your marvelous application. I would recommend using semantic versioning (tagging releases like `1.0.0`, `1.0.1` and so). I've noticed the releases tab on Github is empty, and images are tagged with the date they are released on `ghcr.io`. I personally tend to avoid deploying `latest` since sometimes breaking versions can temporarily end up in tagged like that.

[Semantic Versioning is a standard in the industry](https://semver.org/), for the developers you get the benefit of categorizing every release as bugfix, new feature or a breaking change avoiding potential headaches in the future.

For users, this allows to have more control on which version we deploy and when and what to upgrade.

Is there any reason not to tag releases like this? I'm happy to help if needed with it.
Thanks!

alexta69:
Releases are tagged with the date now.

PR #735, opened 2025-07-30
Change version / tag format to follow conventions

AlexGustafsson:
In order to allow for automated tooling to identify new releases of the
container image, follow best practices by adhering to the semver
specification (at least in terms of syntax).

Solves: #487

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
       -
         name: Get current date
         id: date
-        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+        run: echo "::set-output name=date::$(date +'%Y.%m.%d')"
       -
         name: Checkout
         uses: actions/checkout@v4


AlexGustafsson:
Optionally, one could tag the image using an actual semantic version, but that would require a bit more work to automate fully (as in bumping a patch on dependency upgrades etc.).

alexta69:
But the date has nothing to do with semver -- it happens to have three numbers, but that doesn't make it semver (as you indeed note in your comment). So what's the point of this change? What automated tooling are we talking about here?

AlexGustafsson:
No, exactly. It's not semantic, but it does adhere to the same syntax used by semver. Although semver is probably the most widely used tag format by far, due to this property there are a few projects that have adapted various forms of date formatting. Projects like home assistant, yt-dlp, ubuntu (although only the year).

These projects all have one thing in common, they still follow semver's syntax to integrate with established conventions. As `-` in semver denotes the start of the "trailer", the version currently in use will be interpreted as only the year with the rest of the date being some label. A `.` will mean that the date, although not semantic, will adhere to the syntax and basic notion of "split values by period, compare, higher is newer".

Example of tooling that uses semver to identify new updates:

- Dependabot
- Renovate

(There's also Cupdate, but I wrote it, so let's leave it out for now).

alexta69:
You've convinced me :)
By the way, how does Cupdate compare to watchtower? The latter isn't being updated for a while now, but besides that, is Cupdate a continuation of the same idea, or is it very different?

AlexGustafsson:
@alexta69 I haven't used watchtower, but from what I can tell the big difference would be that Cupdate doesn't intend to change anything in the environment it's running (i.e. carry out the update), which watchtower does. Cupdate's more geared towards answering "what's in use", "what's been updated" and "what's changed in those updates". There's also native support for Kubernetes. Other than that, there's quite a few security-related features that I think are quite unique. Take a look at the demo or try it out! It's linked in the README: https://github.com/AlexGustafsson/cupdate.

PR #740, opened 2025-08-02
Add UI and backend support for download configuration

JakePeralta7:
Introduces UI controls to set download mode and max concurrent downloads, with backend API to update these settings dynamically. The configuration is now synchronized between frontend and backend, allowing users to adjust download scheduling directly from the UI.


alexta69:
The problem I have with this is that the changes from the UI are lost next time MeTube is restarted. So it creates inconsistent behavior -- when you load MeTube, the options may be the ones you set in the UI last time you used it, and may not be, depending on whether there was a restart.

Generally, the guideline is that advanced options are sent with the "add" request, and are valid only for that request, and are not persisted in the backend and don't change its state / configuration in any way. I want to keep it that way. So for this functionality to be added to MeTube, please change it to behave according to that principle.

Thanks for your contribution!

alexta69:
Closing as part of a stale-PR cleanup — the concern about UI-changed settings being lost on restart (inconsistent behavior) was never addressed, and the branch now conflicts with master.

PR #750, opened 2025-08-19
Add "Format" column to frontend interface to display downloaded audio/video formats

xerdream:
image
Metube offers multiple download format options, but the frontend download completion list only allows one format of the same video to exist at a time. Therefore, I modified the code to enable multiple formats of the same video to coexist in the list.

Additionally, I fixed the following issues:

Incomplete temporary file cleanup – Ensured temporary files are properly deleted.

Accidental deletion of previously saved files – Resolved a bug introduced by the changes that could incorrectly remove existing downloaded files.

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -210,7 +210,7 @@ class Download:
         if not os.path.isdir(self.download_dir):
             return
 
-        tmpfilename = self.tmpfilename.rsplit('.')[0]
+        tmpfilename = os.path.basename(self.tmpfilename)
         def is_tmpfile(filename):
             return filename.startswith(tmpfilename)
 


alexta69:
@xerdream Can you please take a look at #756? It's an error that is caused by this change.

xerdream:
> @xerdream Can you please take a look at #756? It's an error that is caused by this change.

OK. I'll deal with it as soon as possible.

alexta69:
#756 has an additional comment, seems like it's not fully fixed. I decided to revert this feature. Sorry about that. I think that the renaming logic is complicated and error-prone. If we're ever to implement such a feature, we should think of a cleaner way that wouldn't introduce so much logic and complexity.

xerdream:
I apologize for the trouble. This renaming approach was meant to resolve an issue caused by the feature, but I can’t think of a more suitable alternative for now. Perhaps it’s best to revert the entire feature. I’ve learned a lot from this.

alexta69:
No worries. Thanks for participating in the project!

PR #751, opened 2025-08-22
run `pipenv lock` to regenerate lock file

jkrehm:
I believe this will fix the [build failure](https://github.com/alexta69/metube/actions/runs/17113355051/job/48539367929).

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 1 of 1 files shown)


evrial:
what this colorama garbage doing at all?

jkrehm:
No idea. I'm not a Python developer and don't know how pipenv deals with dependencies, but it appeared when I ran `pipenv lock`, so I would guess it is a transitive dependency of one of the packages. 🤷

PR #786, opened 2025-09-27
Improve performance of restoring queued/pending items on restart

exterrestris:
When restoring queued/pending items on restart add the saved details directly to the queue rather than sending them through yt-dlp to re-extract the information we already have. Queues should therefore now load virtually instantaneously.

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -31,7 +31,7 @@ class DownloadQueueNotifier:
         raise NotImplementedError
 
 class DownloadInfo:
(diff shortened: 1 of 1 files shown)


exterrestris:
This has the added benefit for YT downloads of not requesting the information for a large number of videos from YT in quick succession at startup, which could trigger 'sign in to prove you're not a bot' (or similar) behaviour thereby blocking downloads until that block is cleared

alexta69:
Thanks @exterrestris !

Issue #495, opened 2024-08-17
[Feature Request]: Rename file


martinikolaev:
As a user I want to be able to rename the files from the UI because some files are named with random GUIDs when I download files in bulk and its hard to rename them later.

Example:
<img width="683" alt="image" src="https://github.com/user-attachments/assets/67b2dfff-6398-4f1d-86e7-753394cdf63a">

StianWiu:
I would love this aswell

PikuZheng:
you can add a  name prefix before add it to download list

celouw:
The rename feature would be great and helpful.

lovejohal:
A text field for the filename would be useful. If the field is left blank, use the default title; otherwise, use the filename entered by the user before adding it to the download.

DaKheera47:
i've added a PR for this (#796). lemme know if that addresses what everyone wanted, or if there's something else that y'all wanted for this feature

Bjoernsen:
Because @DaKheera47 changes have been rejected, I try another PR (#805)

alexta69:
Closing per the project's scope line (see #1027): MeTube writes the file once, as correctly as it can — managing files after they're written (renaming included) is left to file managers and media tools, which is also why the two PRs for this weren't merged. The in-scope way to get this outcome is naming the file correctly *at download time*: the custom-prefix field exists today, and #56 (an optional filename field on the add form) is the open request tracking exactly that — please add your 👍 and use cases there. Closing this one.

PR #796, opened 2025-10-14
Add support for renaming files before and after downloading

DaKheera47:
Renaming can happen before adding a file, and after downloading. the filename on disk is also changed. Related to issue #495 

image

image

image

### DIFF
--- a/ui/src/app/app.component.html
+++ b/ui/src/app/app.component.html
@@ -73,17 +73,17 @@
       <div class="row mb-4">
         <div class="col">
           <div class="input-group input-group-lg shadow-sm">
-            <input type="text"
-                   autocomplete="off"
-                   spellcheck="false"
-                   class="form-control form-control-lg"
-                   placeholder="Enter video or playlist URL"
-                   name="addUrl"
-                   [(ngModel)]="addUrl"
+            <input type="text" 
+                   autocomplete="off" 
+                   spellcheck="false" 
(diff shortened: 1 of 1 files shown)


alexta69:
This is an insane amount of code and repetitious logic for such a simple feature. I assume it was written by an AI agent with little or no oversight. I'm sorry, I cannot accept this PR.

alexta69:
The problem was never with whitespace, but with the actual added code. There's an inordinate amount of logic both on the frontend and the backend to support this. I don't think that it's that important to allow renaming the file in MeTube, when one could just as well rename the file in the filesystem to which the file was downloaded. MeTube isn't designed to provide a full file management solution, that's not its purpose. And the fact that the feature doesn't work when downloading playlists also doesn't fit well with MeTube. We have `OUTPUT_TEMPLATE` and the Custom Name Prefix that work with individual downloads as well as playlists, and I don't see a need for anything else beyond that, especially if it causes such an inflation of code (with the accompanying maintenance burden). I'm sorry I'm rejecting your honest contribution, I don't mean to be mean about it, but it's my job as the maintainer to try to balance code size & complexity with useful functionality, and in this case the addition is not justified in my judgement.

DaKheera47:
That's okay, thanks for looking at it!

The issue discussing this was open (and had been for a while), so I thought it was a valid issue to attempt to resolve

PR #845, opened 2025-12-13
chore(docker): remove deno

pierrenedelec:
Motivation : 

Removing deno save nearly 100mb on docker image.
BE is using python , FE is using Angular , i don't see a reason to have deno . 

with deno : 
Capture d’écran 2025-12-13 à 13 51 33

without deno : 
Capture d’écran 2025-12-13 à 14 05 30


alexta69:
Deno was added for a reason. Check the commit message for that change and the issue it closes.

PR #846, opened 2025-12-14
Switching to a maintained fork of watchtower

aloki:
The original repository has not been maintained for a long time and has problems with newer versions of Docker.


PikuZheng (approved):
nice job

PR #848, opened 2025-12-15
Convert files to LF line ending

alemonmk:
Hi,

I'm repackaging the container for my own, tried making some patch by editing and `git diff > changes.patch`, then found out patching was failing. I saw some weird "^M" in `changes.patch`, turns out some files are committed in CRLF. Hope this PR fixes it.


alexta69:
I've merged another PR that conflicts with this now. In any case, I don't think it should really matter...

alemonmk:
OK, rebased.

> In any case, I don't think it should really matter...

I do stuff under Windows but I never commit CRLF. Honestly it doesn't mattered to me after my next PR...

PR #855, opened 2025-12-26
Suppress download progress update in logs

alemonmk:
Thanks for pulling my previous commit.

This PR aimed to reduce logs when downloading video by sending status update log to debug so it no longer spam the hell out of syslog/journal in regular use. For large enough video these can account for tens of thousands lines of update. (BLAST Rivals Hong Kong Grand Final, 27GB, generated ~75K lines)

image


alexta69:
Thanks @alemonmk !

PR #857, opened 2025-12-27
fix completed result tooltip

mercury233:
**Before the fix:**
GIF 2025-12-27 10-53-33
Items in the completed task list always had a tooltip, even when there was no `msg` or `error`.

**After the fix:**
The tooltip is shown only when there is a `msg` or an `error`.


alexta69:
Thanks @mercury233 !

PR #858, opened 2025-12-30
Improves logging, helpful when debugging yt-dlp options

ikatkov:
I started messing with and yt-dlp options and nothing was working. I set logging to DEBUG in launch.json but log.info() in the main.py does not print anything.
Eventually I figured it all out. This is a PR to help the next person.

## Summary
Improves logging configuration to prevent dropped early log messages and makes yt-dlp verbosity respect the application's log level.

### Early Logging Initialization
- Moved `parseLogLevel()` function to the top of `main.py` for early access
- Configure logging with `basicConfig()` before `Config()` instantiation to capture early startup messages
- Added safety check to avoid overwriting existing handlers when MeTube is embedded in other applications

### Synchronized Log Level Management
- Re-apply log level after `Config()` loads to respect config file settings that may override environment variables
- Ensures Config remains the single source of truth for LOGLEVEL throughout the application

### yt-dlp Verbosity Control
- Made yt-dlp respect DEBUG log level by dynamically checking if DEBUG logging is enabled
- When DEBUG is active, yt-dlp runs with `quiet=False` and `verbose=True`
- Applied to both download operations (`_download()`) and info extraction (`__extract_info()`)
- Production behavior unchanged: yt-dlp remains quiet at INFO and higher log levels

## Benefits
- **No more dropped messages**: Early configuration ensures startup logs are captured
- **Better debugging**: yt-dlp verbose output available when `LOGLEVEL=DEBUG`
- **Cleaner production logs**: yt-dlp remains quiet at standard log levels
- **Consistent behavior**: Log level changes are respected throughout the application lifecycle

## Testing
Tested with:
- `LOGLEVEL=DEBUG` - confirms yt-dlp verbose output appears
- `LOGLEVEL=INFO` - confirms yt-dlp remains quiet (current behavior)
- Various log level configurations to ensure proper fallback handling

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -21,6 +21,26 @@ from yt_dlp.version import __version__ as yt_dlp_version
 
 log = logging.getLogger('main')
 
+def parseLogLevel(logLevel):
+    match logLevel:
+        case 'DEBUG':
+            return logging.DEBUG
+        case 'INFO':
+            return logging.INFO
+        case 'WARNING':
+            return logging.WARNING
+        case 'ERROR':
+            return logging.ERROR
+        case 'CRITICAL':
+            return logging.CRITICAL
+        case _:
+            return None
+
+# Configure logging before Config() uses it so early messages are not dropped.
+# Only configure if no handlers are set (avoid clobbering hosting app settings).
+if not logging.getLogger().hasHandlers():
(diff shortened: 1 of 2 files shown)


alexta69:
Thanks @ikatkov !

Issue #859, opened 2025-12-31
[Bug]: KeyError: '<URL>'


AlvinRamoutar:
### Prerequisites

- [x] I have searched existing issues and discussions to ensure this bug hasn't been reported before
- [x] I have read the [troubleshooting section](https://github.com/alexta69/metube#-troubleshooting-and-submitting-issues) in the README
- [x] I have tested this issue with yt-dlp directly (not just through MeTube UI) as described in the README
- [x] I have checked that this is not a yt-dlp issue (if it is, please report it to [yt-dlp repository](https://github.com/yt-dlp/yt-dlp/issues) instead)

### Bug Description

When restarting metube, it fails to load the `completed` gdbm.

*Expected:* Restarting metube should restart.

*Actual:* Fails to start with the following log:
```
Setting umask to 022
Creating download directory (/downloads), state directory (/downloads/.metube), and temp dir (/downloads)
User set by docker; running MeTube as 1000:0
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/shelve.py", line 111, in __getitem__
    value = self.cache[key]
            ~~~~~~~~~~^^^^^
KeyError: 'www.youtube.com/watch?v=nNQSzsql3d0'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/app/main.py", line 150, in <module>
    dqueue = DownloadQueue(config, Notifier())
  File "/app/app/ytdl.py", line 239, in __init__
    self.done.load()
    ~~~~~~~~~~~~~~^^
  File "/app/app/ytdl.py", line 189, in load
    for k, v in self.saved_items():
                ~~~~~~~~~~~~~~~~^^
  File "/app/app/ytdl.py", line 203, in saved_items
    return sorted(shelf.items(), key=lambda item: item[1].timestamp)
  File "<frozen _collections_abc>", line 899, in __iter__
  File "/usr/local/lib/python3.13/shelve.py", line 113, in __getitem__
    f = BytesIO(self.dict[key.encode(self.keyencoding)])
                ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: b'www.youtube.com/watch?v=nNQSzsql3d0'
```

### yt-dlp Direct Test Results

n/a

### Environment & Configuration

Ubuntu 24.04.3 LTS, metube:2025.12.30

PikuZheng:
[...]

PR #862, opened 2025-12-31
feature/repair-persistent-queues

AlvinRamoutar:
Adding support to repair persistent queues (REF #859 )
- fail-open (if we cannot repair, proceed)
- repair is debug-level (can change, thought it wasn't necessary to pollute logs since the intent is to happen every startup)
- support gdbm and sqlite3 since newer versions of shelve now use sqlite3 as the default backend (see https://discuss.python.org/t/new-default-preferred-dbm-backend/44228)
- bundle gdbmtool and sqlite3 cli into container image

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import yt_dlp
 from collections import OrderedDict
 import shelve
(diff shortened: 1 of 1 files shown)


alexta69:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/alexta69/metube/pull/862#pullrequestreview-3622294866)

AlvinRamoutar:
Thanks @alexta69 - went ahead and addressed feedback with the latest commit.

alexta69:
Thanks @AlvinRamoutar !

PR #877, opened 2026-01-13
Add environment variable to skip changing ownership of directories on startup

### LINKED ISSUE #628: Disable "Changing ownership of download and state directories"
My download directory is massive, every time I start the container I have to go through a very lengthy chown command.

Ownership has never been an issue for me. There should be an option to disable this functionality so it doesn't execute every start.

its-wizza:
I use an external mount as my `/downloads` path with permissions set on the outside of the local filesystem for my `UID`/`GID`. Therefore I get many `Operation not permitted` errors on container startup for the download/state/temp dirs.

This pull request allows for the setting of an additional environment variable `CHOWN_DIRS` to `false` to avoid this on startup.

Should also speed up startup for those with large download directories.

Fixes #628


alexta69:
Thanks @its-wizza and sorry for the delay.

PR #893, opened 2026-02-07
add initial docker-compose configuration for metube service

forzayt:
The docker compose file was missing which cause the crashing of "docker composer up"  command to run


alexta69:
I don't see a reason for this, since it's not a working file anyway (with "/path/to/downloads" in it). There's an example in the README, that's sufficient

PR #904, opened 2026-02-21
Captions extraction feature with advanced options

vitaliibudnyi:
In the video format selector there's a new "Captions" option.
In the Advanced Options Panel there are 3 new selectors: subtitles file format, language and source.

File formats: .srt, .vtt, .ttml, .txt;
Languages: English, Spanish, French, German, Italian, Portuguese, Russian, Ukrainian, Japanese, Korean, Chinese (Simplified), maybe worth adding more;
Sources: manual only, automatically generated only, prefer manual, prefer automatically generated (last one might be redundant, but only experienced YouTube users can tell, not me).

.txt format created by stripping time stamps and cues from intermediate .srt

Known issues: it downloads both auto generated and manually written subs, due to how yt-dl works, better solution would require multiple steps (multiple yt-dl passes). It seems tolerable as subs are usually only hundreds of KB in size. Also I never debugged with DELETE_FILE_ON_TRASHCAN=true flag, which AFAIK probably would cause issues because multiple files per download might not be tracked correctly and therefore not deleted upon request.

In general this is a more complex approach, in case someone wants to add subs embedding in the future. Otherwise it might make more sense to have no new stuff in the Advanced Options Panel at all. And to simply have "Captions" format, which would download autogenerated .srt and convert that to .txt for simplicity. But it's very limited.

Common use case for this feature: to summarize long videos from subscription feed, before watching, saves a lot of time. Also low-effort research is much easier and faster when you do search on text on your own drive, instead of skipping through videos and scrolling through sponsored search results online. Same problem: lot's of valuable info on YouTube, but lots of trash as well.


alexta69:
Thanks @vitaliibudnyi !

PR #907, opened 2026-02-24
Use youtube time parameters in the URL query string as the starting point for yt-dlp

hinmanjp:
If a url is provided with a query string that includes t=855s (or similar), the time value will be parsed out and passed to yt-dlp as the starting time.

Also allow manual builds under github actions.


alexta69:
@hinmanjp can you please rebase on master?

Issue #610, opened 2025-03-10
Sort downloads by most recent first


Mladia:
I'd like to request a feature to enable sorting the downloaded files in the Completed section so I can see the most recent files first. I have been using the application for a long time, and I recall that this was the default behavior initially. However, at some point, it changed.

EvgeniyDoctor:
+1.

PR #911, opened 2026-03-01
Sort completed downloads by newest first

ddmoney420:
Addresses #610.

Adds a sort toggle to the Completed section — defaults to newest first, click to flip. Preference persists via cookie.

Frontend-only change, no backend modifications.

**What changed:**
- `app.ts`: `sortedDone()` method sorts by `timestamp`, `toggleSortOrder()` saves preference
- `app.html`: sort button in Completed header, done table uses `sortedDone()` instead of raw map iteration

Happy to adjust if anything doesn't match the project's conventions.

### DIFF
--- a/ui/src/app/app.html
+++ b/ui/src/app/app.html
@@ -424,7 +424,14 @@
     </table>
   </div>
 
-  <div class="metube-section-header">Completed</div>
+  <div class="metube-section-header d-flex align-items-center">
+    Completed
+    <button type="button" class="btn btn-sm btn-outline-secondary ms-3"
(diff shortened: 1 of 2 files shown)


alexta69:
You have submitted three PRs based on master, that touch the same files, so they can't all be merged, they create conflicts.
I understand the desire to separate the fixes for the different issues, but the way to do that is to create separate commits for them, one on top of the other, all submitted in the same PR. Then we can review them and merge the PR.
Thanks

ddmoney420:
Combining into a single PR per Alex's feedback — separate commits, one PR.

Issue #143, opened 2022-05-19
Shoe errors in the web interface


1RandomDev:
There are many cases which happen frequently during normal use where yt-dlp gives an error. (e.g. age restricted video, already recorded in archive, TikToks with so many hashtags that it exceeds the maximum length for file names, characters in the name that are illegal in file names, etc.)
So it would be extremely helpful if any errors yt-dlp gives are forwarded to the webinterface instead of just saying "Server got itself in trouble".
Oh and it seems like not even in the logs you can see those error messages.

alexta69:
Yes, I agree it would be a useful addition.

Verkhovskyi:
Still happens. Spent some time trying to understand why can't MeTube download the video. It was age restricted

PR #912, opened 2026-03-01
Expandable error details with copy-to-clipboard

ddmoney420:
Addresses #143.

Failed downloads now have a clickable error icon that expands to show full error details — message, error text, and URL. There's a copy button so users can easily paste into bug reports.

Frontend-only change, no backend modifications.

**What changed:**
- `app.ts`: `toggleErrorDetail()`, `copyErrorMessage()`, `isErrorExpanded()` methods + `expandedErrors` set
- `app.html`: error icon is clickable, expands an alert panel with details and a copy button

Tried to keep it consistent with the existing error display style.

### DIFF
--- a/ui/src/app/app.html
+++ b/ui/src/app/app.html
@@ -456,20 +456,42 @@
                   <fa-icon [icon]="faCheckCircle" class="text-success" />
                 }
                 @if (download.value.status === 'error') {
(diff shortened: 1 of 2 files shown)


ddmoney420:
Combining into a single PR per Alex's feedback.

Issue #840, opened 2025-12-07
[Feature]: Ability to cancel playlist adding


ThePoetics:
### Prerequisites

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before
- [x] I have read the [feature request guidelines](https://github.com/alexta69/metube#-submitting-feature-requests) in the README
- [x] I understand that MeTube development relies on community contributions and the maintainer is not likely to implement this feature
- [x] I have considered if this feature should be implemented in yt-dlp instead of MeTube (if so, please report to [yt-dlp repository](https://github.com/yt-dlp/yt-dlp/issues))

### Feature Description

Currently there is no way to stop the automatic parsing and addition of playlist entries once started (say, if someone intended to add a single video but linked to the playlist, each successive video is added).

It would be very nice to have a button to stop the processor from seeking out more entries.

### Implementation Plan

No, I'm requesting this feature but cannot implement it

seanwlk:
Just enable strict playlist mode. You can turn it on by default from env
`DEFAULT_OPTION_PLAYLIST_STRICT_MODE=true`
This way if someone pastes a link from a playlist, it won't download everything, you'll need an actual playlist link to trigger the full parsing

ThePoetics:
Thank you; I hadn't seen that option before and will be making use of it in the future.

Is it still worthwhile to have an option to stop parsing a playlist, however? I could see cases where someone does intentionally enter a playlist URL but doesn't expect the playlist to contain 500 entries or something similar.

seanwlk:
With strict mode you significantly lower the chances of errors. The option would be indeed useful in some cases but in the end it's just a matter of whom you share the app with. 
In my case i trust that the few people i exposed the app to, are smart enough to not go and do that. Also considering that you have to intentionally go and copy the playlist link and not the music / video link.

I only found out about the strict mode because I myself pasted a couple of times the link of a song that was part of a playlist and ended up making a mess, I actually think this should be on by default.

PR #913, opened 2026-03-01
Cancel playlist adding mid-operation

ddmoney420:
Addresses #840.

Adds a cancel button that appears while a playlist is being added. Clicking it sets an abort flag that's checked between entries in the playlist loop, so already-queued items stay and the rest are skipped.

**What changed:**
- `ytdl.py`: `_add_canceled` flag, `cancel_add()` method, flag check at top of entry loop, reset on fresh `add()` calls
- `main.py`: `POST /cancel-add` route + CORS
- `downloads.service.ts`: `cancelAdd()` service method
- `app.ts`: `cancelAdding()` method
- `app.html`: download button swaps to red Cancel button when `addInProgress`

Kept the backend changes minimal — just a boolean flag and one route.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -322,6 +322,11 @@ async def start(request):
     status = await dqueue.start_pending(ids)
     return web.Response(text=serializer.encode(status))
 
+@routes.post(config.URL_PREFIX + 'cancel-add')
+async def cancel_add(request):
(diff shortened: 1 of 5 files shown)


ddmoney420:
Combining into a single PR per Alex's feedback.

PR #918, opened 2026-03-02
Fix deleted playlist videos being re-queued

ddmoney420:
When adding a big playlist and deleting individual videos from the queue mid-add, they come right back. The playlist loop checks `queue.exists(key)` before adding — once you delete something, it's no longer in the queue, so the loop re-adds it.

Fix: track canceled URLs in a set that `__add_entry` checks before re-queuing. Set clears on each fresh `add()` call so it doesn't leak.

Found this while testing the cancel feature from #915.


alexta69:
Thanks @ddmoney420 !

PR #926, opened 2026-03-09
Add video codec selector and codec/quality columns in done list

cycl0ne:
## Summary

   - Adds a **Video Codec** dropdown (Auto, H.264, H.265/HEVC, AV1, VP9) to the download form, visible when downloading video formats
   - When a codec is selected, yt-dlp format strings are filtered to prefer that codec, with automatic fallback to the best available format if the preferred codec isn't found
   - Adds **Quality** and **Codec** columns to the completed downloads table for at-a-glance info on what was downloaded
   - Codec preference is persisted in a browser cookie across sessions

   ## How it works

   The codec selector builds yt-dlp format filter strings (e.g. `[vcodec~='^(h264|avc)']`) that are prepended to the format selection. If no matching format exists, yt-dlp falls back to the unfiltered best format — no download will fail due to
   codec preference.

   No re-encoding or FFmpeg postprocessing is involved; this purely selects from natively available formats. (Was thinking of doing a reencode option for ffmpeg, but i discarded it)

### DIFF
--- a/app/dl_formats.py
+++ b/app/dl_formats.py
@@ -3,6 +3,13 @@ import copy
 AUDIO_FORMATS = ("m4a", "mp3", "opus", "wav", "flac")
 CAPTION_MODES = ("auto_only", "manual_only", "prefer_manual", "prefer_auto")
 
+CODEC_FILTER_MAP = {
+    'h264': "[vcodec~='^(h264|avc)']",
+    'h265': "[vcodec~='^(h265|hevc)']",
+    'av1':  "[vcodec~='^av0?1']",
+    'vp9':  "[vcodec~='^vp0?9']",
+}
+
 
 def _normalize_caption_mode(mode: str) -> str:
     mode = (mode or "").strip()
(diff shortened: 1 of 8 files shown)


alexta69:
We already have a format dropdown that includes codecs (mp4, mp3, and so on). A codec dropdown like the one you're describing sounds an awful lot like what we already have. What is the rationale for this addition?

cycl0ne:
First of all this is hidden under Advanced Options, to not confuse people
Second you only have mp4, mp3 etc as Container Format. But here its the video codec/compression format. Most videos you get from youtube are now vp9, but whatsapp only takes h23x as example. With this you can now choose which compression (codec) you want and can send this video to whatsapp as an example. (thats in my case, i hate sending links to youtube short videos, i send directly the 4mb video file).

And the other patch in this code: 
you can see which codec/container... this is also now added to the download list.

alexta69:
I ended up restructuring the whole thing, I think it's organized in a way that makes much more sense now. Thanks for kicking this off. :)

cycl0ne:
Perfect! Makes sense.

didnt wanted to restructure the code too much, since its not mine. ;-)

PR #935, opened 2026-03-31
Refactor app bootstrap and harden container configuration

jalvarado-it:
Extract environment-backed settings into a dedicated module and switch the aiohttp app to a factory-based bootstrap.

Move runtime dependencies into typed app state, tighten CORS defaults, reject sensitive inline yt-dlp options, and harden the container entrypoint validation and ownership flow.

Update tests to cover settings validation, app factory behavior, CORS policy, and entrypoint safety checks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ pending*
 
 __pycache__
 .venv
+
+# Testing
+./local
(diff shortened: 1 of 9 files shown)


alexta69:
There are lots of unnecessary changes here (variable and function renaming etc.) and it's quite hard to review. I can't merge this PR, sorry.

If there are specific changes that you want to suggest, please create a PR with only these changes. Let's leave out the housekeeping.

Thanks for participating in the project!

PR #941, opened 2026-04-03
Feature: Optional completion webhook + configurable UI refresh throttling

Stash-KennyG:
### 1. Optional completion webhook

Adds an **optional, generic HTTP webhook** that runs after each completed download, posting a **fixed JSON payload** to a user-configured URL. Configuration is available via **environment variables**, the **web UI** (Advanced Options), and **REST endpoints**. Settings changed through the UI are **persisted under `STATE_DIR`** so they survive container restarts.

Also includes a **Pretty / Raw** toggle when exercising the webhook **Test Service** response.

### 2. Configurable UI refresh throttling (“gear” settings)

**Stock mainline MeTube** fires an Angular `updated` event on **every** Socket.IO download `updated` tick so the queue UI stays maximally fresh. That is unchanged for users who never touch the gear.

**This PR** adds an *optional* code path to **coalesce** those events so the UI repaints at most once per configured interval (separate values for focused vs background tab). That helps when the same “every tick” behavior drives **high CPU** on large queues or weaker clients—the tradeoff the fork originally hit before coalescing existed.

The **navbar gear** dialog sets **Focused refresh (ms)** and **Background refresh (ms)** (browser cookies). **Defaults are `0` / `0`**, which disables coalescing and preserves mainline’s every-tick behavior. Non-zero values turn on coalescing for users who want it.

Webhook default behavior is unchanged when disabled (`WEBSERVICE_ENABLED=false`).

---

## Motivation

**Webhook:** Downstream automation often needs a simple “file landed on disk” signal without tying MeTube to a specific consumer. A **single POST URL** and **stable JSON shape** keep integrations portable (media libraries, custom scripts, arbitrary services). Persisting UI-saved settings avoids losing configuration on restart when using Docker volumes for `STATE_DIR`.

**Refresh throttling:** Mainline ties UI refresh to **every** progress socket message. A fork that kept that model verbatim saw **sustained CPU** (and mobile battery impact) with busy queues. Optional coalescing addresses that **without changing defaults**: **`0` ms = same as mainline**; larger intervals are opt-in.

---

## User-visible changes

- **Advanced Options → Webhook integration:** enable/disable, endpoint URL, timeout, **Save Webhook**, **Test Service**.
- **Test Service** calls the configured endpoint with a **validation-style body** (empty `filePath`); response is shown as a **key/value table (Pretty)** or **formatted JSON (Raw)**.
- **Navbar → gear icon:** **Focused refresh (ms)** and **Background refresh (ms)** — `0` = stock behavior (every progress event); larger values coalesce updates (cookie-backed; clamped in code).

---

### Webhook (server / env)

| Variable | Purpose |
|----------|---------|
| `WEBSERVICE_ENABLED` | `true` / `false` (default `false`) |
| `WEBSERVICE_ENDPOINT` | Full URL for `POST` (default placeholder in tree) |
| `WEBSERVICE_TIMEOUT_SEC` | Per-request timeout (seconds) |

If `{STATE_DIR}/webhook_settings.json` exists (after a save from the UI), its contents override the above env defaults for the running process.

### UI refresh (client / cookies)

Persisted in the browser (not on the server):

| Cookie | Purpose |
|--------|---------|
| `metube_focused_refresh_ms` | Min delay before coalescing when the tab is active; **`0` = no coalescing (mainline parity)**. Upper bound 10 s when &gt; 0. |
| `metube_background_refresh_ms` | Same when `document.hidden`; **`0` = no coalescing**. Upper bound 120 s when &gt; 0. |

If cookies are absent, both stay **0** (stock behavior).

---

## API (relative to `URL_PREFIX`)

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `webhook-settings` | Current effective settings |
| `POST` | `webhook-settings` | Update settings (and persist when possible) |
| `POST` | `webhook-settings/test` | POST stub payload to the endpoint; return status + parsed body |

---

## Completion payload (fixed shape)

For each completed output file, when enabled:

```json
{
  "filePath": "/path/inside/container/to/file.ext",
  "source": "metube",
  "jobId": "<download id>"
}
```

The test route uses the same shape with `filePath: ""` so consumers can treat it as a **connectivity / validation** request if they choose.

---

## Technical notes (for reviewers)

- **Backend:** `aiohttp` routes; completion hook posts via `urllib` in a thread pool to avoid blocking the event loop.
- **Persistence:** `AtomicJsonStore` (`state_store.py`) writes `webhook_settings.json` under `STATE_DIR` on successful `POST /webhook-settings`; load on startup.
- **Frontend:** `DownloadsService` still subscribes to Socket.IO `updated` the same way as mainline; when both configured intervals are **0**, it calls `updated.next()` on **every** event (identical to mainline). With a **positive** interval it coalesces. The gear modal calls `setRefreshCadence`. No extra Rx `auditTime` on `updated`—that would diverge from mainline when interval is 0.

---

## Testing

**Webhook**

- [ ] Manual: webhook disabled → no extra HTTP traffic on completion.
- [ ] Manual: enable + point to a local echo/HTTPbin → completion POST received with expected JSON fields.
- [ ] Manual: save settings in UI → restart container → settings still applied (volume on `STATE_DIR`).
- [ ] `POST /webhook-settings/test` → Pretty and Raw views render.

**Refresh throttling**

- [ ] Default / no cookies / both **0** → progress updates match mainline (Perceived every-tick repaint).
- [ ] Set focused interval **e.g. 3000** → CPU drops vs **0** on a busy queue.
- [ ] Reload page → cookie values still applied.
- [ ] Hidden tab: non-zero **background** interval uses `document.hidden` path.

**General**

- [ ] CI / existing test suite passes.

---

## Scope / non-goals

- Does not change yt-dlp download behavior beyond an optional post-completion HTTP call.
- Does not prescribe a particular webhook consumer; endpoint-specific behavior is out of scope.
- UI refresh cookies are **per browser**; not synced across devices.


alexta69:
This PR contains a massive amount of changes, including github workflow changes that are irrelevant, housekeeping of various kinds, and functionality that is not needed (more on that below). I cannot accept the PR, I'm sorry. I'm closing it. I'm thankful for participation in the project, but if PRs are to be accepted, they have to only contain concise changes that are meaningful for one fix or feature, and we have to agree that the fix / feature is indeed necessary.

Regarding the functionality:

Webhooks -- it's possible to configure `yt-dlp` to execute a command upon completion (via `postprocessors`), and that can be used to invoke `curl`. I don't see a need for an additional implementation of webhooks in MeTube.

UI refresh -- I'm not convinced that this is necessary, but even if it is, there's no reason to expose this configuration in the UI, it's completely internal and shouldn't concern the user. Let's think if we can arrive at a sensible default here that we can confidently apply to everybody. If we can, perhaps we should implement it.

PR #944, opened 2026-04-05
Fix string boolean parsing for subscription enabled updates

jacinli:
## Summary

Fix subscription updates so the `enabled` field is parsed explicitly instead of relying on Python truthiness.

Previously, `SubscriptionManager.update_subscription()` used `bool(changes["enabled"])`, which incorrectly treated non-empty strings such as `"false"` as `True`. This could cause API clients that send string boolean values to resume a subscription instead of pausing it.

## Changes

- add explicit boolean coercion for subscription `enabled` updates
- accept common string boolean values such as `true` / `false`, `1` / `0`, and `on` / `off`
- reject invalid values instead of silently treating them as truthy
- add regression tests covering `"false"` and invalid input handling

## Testing

```bash
python3 -m pytest -q app/tests/test_subscriptions.py -k 'update_subscription_parses_string_false_enabled or update_subscription_rejects_invalid_enabled_value'
```

### DIFF
--- a/app/subscriptions.py
+++ b/app/subscriptions.py
@@ -231,6 +231,19 @@ def _subscription_from_record(record: Any) -> Optional[SubscriptionInfo]:
     return None
 
 
+def _coerce_bool(value: Any) -> bool:
+    """Accept JSON booleans and common string forms used by API clients."""
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, str):
+        lowered = value.strip().lower()
+        if lowered in {"true", "1", "on"}:
+            return True
+        if lowered in {"false", "0", "off"}:
+            return False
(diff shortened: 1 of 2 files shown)


alexta69:
Thanks @jacinli. I'm curious, in what flow have you encountered this problem? The UI as currently implemented is not triggering the problem, right?

jacinli:
Yes, that's right — I did not hit this through the current UI.

I noticed it while reviewing the backend API. If `enabled` is sent as the string `"false"` instead of a JSON boolean, `bool("false")` evaluates to `True` in Python, so the subscription would be resumed instead of paused.

So this was more of a defensive API fix for non-UI clients and future integrations.

PR #949, opened 2026-04-10
Fix permissive CORS policy that allows cross-origin attacks

az10b:
## Summary

- The `on_prepare` handler unconditionally reflects the `Origin` request header into `Access-Control-Allow-Origin`, and Socket.IO is configured with `cors_allowed_origins='*'`
- This allows **any website** to make cross-origin requests to all API endpoints without authentication
- An attacker can force the server to download arbitrary content (`POST /add`), overwrite cookies (`POST /upload-cookies`), delete downloads (`POST /delete`), and create subscriptions (`POST /subscribe`) — all from a malicious webpage the victim visits

## Fix

- Replace blanket origin reflection with an explicit allowlist via the `CORS_ALLOWED_ORIGINS` environment variable
- When unset, cross-origin requests are denied by default
- Users who need cross-origin access (browser extensions, external clients) can set `CORS_ALLOWED_ORIGINS` to a comma-separated list of trusted origins

## Reproduction

1. Run MeTube: `docker run -d -p 8081:8081 alexta69/metube`
2. Open this HTML file in a browser:

```html
<script>
fetch('http://localhost:8081/add', {
    method: 'POST',
    headers: {'Content-Type': 'application/json'},
    body: JSON.stringify({
        url: 'https://www.youtube.com/watch?v=dQw4w9WgXcQ',
        quality: 'best',
        download_type: 'video',
        format: 'any',
        codec: 'auto'
    })
}).then(r => r.text()).then(console.log);
</script>
```

3. Observe the download appears in MeTube UI — the cross-origin request succeeds

## Test plan

- [ ] Verify MeTube UI still works normally (same-origin, no CORS needed)
- [ ] Verify cross-origin requests are blocked when `CORS_ALLOWED_ORIGINS` is unset
- [ ] Verify cross-origin requests work when `CORS_ALLOWED_ORIGINS` is set to the requesting origin
- [ ] Verify Socket.IO connections respect the same allowlist

PR #950, opened 2026-04-10
Block dangerous yt-dlp options in ytdl_options_overrides

az10b:
## Summary

- When `ALLOW_YTDL_OPTIONS_OVERRIDES=true`, the `/add` endpoint accepts arbitrary yt-dlp options as JSON and passes them directly to `yt_dlp.YoutubeDL()` without restriction
- An attacker can inject a yt-dlp `Exec` postprocessor that executes arbitrary OS commands on the server after a download completes
- Combined with the permissive CORS policy (#949), this is exploitable from any website the victim visits

## Fix

Add a blocklist of dangerous yt-dlp option keys that are rejected when present in overrides:

- `exec_cmd`, `exec` — execute shell commands after download
- `postprocessors`, `post_hooks` — inject arbitrary postprocessors (including `Exec`)
- `external_downloader`, `external_downloader_args` — specify custom download executables
- `cookiefile`, `cookiesfrombrowser` — read arbitrary cookie files from the filesystem

## Reproduction

1. Run MeTube with overrides enabled:
   ```
   docker run -d -p 8081:8081 -e ALLOW_YTDL_OPTIONS_OVERRIDES=true alexta69/metube
   ```

2. Send a request with a malicious postprocessor:
   ```bash
   curl -X POST http://localhost:8081/add \
     -H 'Content-Type: application/json' \
     -d '{
       "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
       "quality": "128",
       "download_type": "audio",
       "format": "mp3",
       "codec": "auto",
       "ytdl_options_overrides": {
         "postprocessors": [{"key": "Exec", "exec_cmd": "touch /tmp/rce-proof"}]
       }
     }'
   ```

3. After the download completes, verify:
   ```
   docker exec <container> ls -la /tmp/rce-proof
   ```

## Test plan

- [ ] Verify `postprocessors` in overrides is rejected with 400
- [ ] Verify `exec_cmd` in overrides is rejected with 400
- [ ] Verify safe overrides (e.g. `{"format": "bestaudio"}`) still work
- [ ] Verify server-side `YTDL_OPTIONS` and `YTDL_OPTIONS_PRESETS` are not affected by the blocklist

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -234,6 +234,12 @@ VALID_VIDEO_CODECS = {'auto', 'h264', 'h265', 'av1', 'vp9'}
 VALID_VIDEO_FORMATS = {'any', 'mp4', 'ios'}
 VALID_AUDIO_FORMATS = {'m4a', 'mp3', 'opus', 'wav', 'flac'}
 VALID_THUMBNAIL_FORMATS = {'jpg'}
+_BLOCKED_YTDL_OVERRIDE_KEYS = frozenset({
+    'exec_cmd', 'exec', 'postprocessors', 'post_hooks',
+    'external_downloader', 'external_downloader_args',
+    'cookiefile', 'cookiesfrombrowser',
+})
+
 def _parse_ytdl_options_overrides(value, *, enabled: bool) -> dict:
     if value is None or value == '':
         return {}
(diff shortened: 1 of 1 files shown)


alexta69:
The overrides option is disabled by default, and must be explicitly enabled by the user. The risk of enabling it is clearly documented. If the user does it anyway, it's their risk analysis, and their decision. Moreover, there might be other dangerous yt-dlp options that are not blacklisted in this PR, and more might be added in that project in the future and not included here, so besides hamstringing the user (who decided to enable this feature perhaps specifically to use these options), we would be providing a false sense of security.

The software should be secure by default (that's why I accepted your CORS PR from today, thanks), but flexible enough to allow diverse use cases for users who make their own risk analysis.

Issue #931, opened 2026-03-21
[Feature] : PASSWORD


Falous-dev:
### Prerequisites

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before
- [x] I have read the [feature request guidelines](https://github.com/alexta69/metube#-submitting-feature-requests) in the README
- [x] I understand that MeTube development relies on community contributions and the maintainer is not likely to implement this feature
- [x] I have considered if this feature should be implemented in yt-dlp instead of MeTube (if so, please report to [yt-dlp repository](https://github.com/yt-dlp/yt-dlp/issues))

### Feature Description

MeTube is currently accessible to anyone who reaches the interface. To improve basic security and prevent unauthorized access, this feature request proposes adding a simple password‑protection layer.
This is not a full authentication system with user accounts — only a single password required to access the application.

Proposed Solution

Introduce a lightweight access gate that displays a login screen when the user is not authenticated.
Access is granted only after entering the correct password, which is defined in the application configuration.

Requirements

• Display a /login page when no valid session is detected.
• Provide a single password input field and a “Submit” button.
• Store the password in a configuration file or environment variable (e.g., METUBE_PASSWORD).
• Use a simple session mechanism (cookie‑based) to remember authenticated users.
• Add a “Logout” option to invalidate the session.
• Ensure the password is never exposed in client‑side code.

Rationale

• Prevent public or unintended access to the MeTube interface.
• Provide a minimal security layer without implementing full user management.
• Keep the solution lightweight and easy to maintain.

Acceptance Criteria

• MeTube cannot be accessed without entering the correct password.
• The session persists correctly and expires when expected.
• Incorrect passwords deny access without revealing sensitive information.
• The implementation is documented for deployment and configuration.

### Implementation Plan

None

leesiuu:
You can also use Traefik or similar tool to make this happen without any change on this package side :)

cubemin:
Definitely want this. Using another layer like Traefik would be overkill, IMHO.

EDIT: Actually, I'm already using NGINX Proxy Manager and managed to set up basic auth for MeTube with that. It's still buggy though - once I log in via the auth layer, there's no way to log OUT. Even if I close and relaunch my browser, going back to the same IP brings the MeTube interface up again right away.

I'd just prefer a native method built into MeTube...

alexta69:
[...]

PR #961, opened 2026-04-19
feat(auth): add authentication system changes

abyssox:
This PR adds optional password protection to MeTube as requested in #931.

## Changes

* add a login page for unauthenticated users
* require a configured password before granting access 
* keep the password on the server side and out of client code
* use a session/cookie-based mechanism for authenticated access
* add logout support

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ services:
       - /path/to/downloads:/downloads
 ```
 
+To require a password for MeTube, set `METUBE_PASSWORD`, or mount a Docker secret and point `METUBE_PASSWORD_FILE` to it, for example `/run/secrets/metube_password`.
+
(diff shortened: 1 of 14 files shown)


alexta69:
Hi @abyssox ,
Thanks for contributing to MeTube. I must reject this PR, because it's an enormous amount of code for such a feature, with a lot of unnecessary complexity. As mentioned in the [Submitting features](https://github.com/alexta69/metube#-submitting-feature-requests) section of the README, it's best to discuss the planned implementation first, most certainly if it adds as much code and complexity as this one.

Besides the above, I'm not convinced about the necessity of password protection in MeTube itself, given how easy it is to add via a simple sidecar like Authelia (or others).

PR #963, opened 2026-04-21
Updated import and fixed race condition

rdiaz738:
Updated the Import button to send all the urls as Promise All requests instead of doing it one by one, added progress bar. Also fixed issue where if you  select all and delete, urls would still be downloading instead of stopping


alexta69:
Thanks @rdiaz738!

PR #977, opened 2026-05-04
make ui more mobile mobile-friendly

adiktofsugar:
I was using this on mobile and couldn't see what I was typing. For some reason the input shows on top of the actual text. It does work, so you see the list being filtered, but you can't see what you've actually written. Also, the tooltip shows on top of the combobox.

Here's what it looks like (I typed "se" in the box. you can see it if you select the text):

#### Current

metube-mobile-screenshot--old

#### Changed

metube-mobile-screenshot--new

### The fixes
To fix the first (main) issue, I replaced the ng-select with an autocomplete.
To fix the second issue, I removed the tooltips and replaced with clickable underlined labels

#### Autocomplete instead of combobox

image

#### Clickable labels

image

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -5,7 +5,7 @@
     "ng": "ng",
     "start": "ng serve",
     "build": "ng build",
-    "build:watch": "ng build --watch",
+    "build:watch": "ng build --watch --configuration development",
     "test": "ng test",
     "lint": "ng lint"
   },
--- a/ui/src/app/app.html
+++ b/ui/src/app/app.html
@@ -279,7 +279,7 @@
           </div>
           <div class="col-12 col-md-6 col-lg-3">
             <div class="input-group">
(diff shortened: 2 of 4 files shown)


alexta69:
I'm not sure about this. The question marks everywhere are a visual nuisance, I'd really rather avoid that. And regarding the picker -- does this mean that you have to start typing the folder name? You can't simply look at the options you have? I understand that the current situation needs fixing, but is eliminating the combo box the right solution?

adiktofsugar:
@alexta69 You can still look at the options, typing just filters the list (I just checked). You can't scroll horizontally, which is kind of annoying, but it's an improvement.

RE: the question marks as a nuisance, we can either do this on mobile only (`@media(hover:none)`) or...move the icons, if you have a spot you prefer (just mark up a screenshot and I'll do my best to get it over there).

alexta69:
I can't think of a good spot for the question marks frankly.. perhaps the tooltip can appear on click? Or not appear at all, if there's no good solution.

adiktofsugar:
@alexta69 Alright, I finally got around to updating this. I removed the icons and made the labels have an underline instead. The last screenshot in the description shows what it looks like. I didn't bother replacing all the others because it takes a while, especially getting it from my phone. But it does work correctly.

PR #986, opened 2026-05-24
Add time-range clips and batch clip downloads

PepegaSan:
https://github.com/alexta69/metube/issues/989
implements batch clips and optional ffmpeg merge (multiple ranges per URL).
Single clip start/end already exists upstream; this PR adds the batch table, `POST /add-batch`, merge, and per-clip queue keys.

MeTube already supports one **Clip start** / **Clip end** per download. This PR adds **batch clips** (several ranges for the same URL) and optional **ffmpeg merge** into a single file.
- **Batch clips** table in Advanced Options: **Download each clip** or **Download merged**
- `POST /add-batch` API; optional `merge_clips`
- Per-clip `queue_key` so multiple clips on the same URL can be queued, cleared, and deleted independently
- Automatic filename prefixes (`clip_…`, `clipbatch_…`, `clipmerged_…`)
- Optional URL import via `?url=…&clips=…` / `#mt=…` for bookmarklets or external tools (no bundled extension in this PR)

## Motivation
A single clip range is already covered by the existing UI. This fills the gap for multiple segments from one video—either as separate files or one merged output—without manual re-queueing or external editing.
## UI
- Existing **Clip start** / **Clip end** unchanged for one range
- New **Batch clips** table: multiple start/end rows, **Download each clip** or **Download merged**
- Optional URL import via `?url=…&clips=…` / `#mt=…`
## API
- `POST /add-batch` — `{ "url", "clips": [{"start","end"}, …], "merge_clips": false, …same fields as /add }`
- Single-clip `/add` behaviour improved where needed (e.g. distinct `queue_key`, filename prefixes)
## Tests
- `app/tests/test_clip_queue_keys.py`
- `app/tests/test_batch_clips.py`
- Updates in `app/tests/test_download_queue.py`
- UI: `ui/src/app/metube-import-query.spec.ts`, `downloads.service.spec.ts`
## Test plan
- [ ] `cd app && pytest tests/test_clip_queue_keys.py tests/test_batch_clips.py tests/test_download_queue.py`
- [ ] Batch: two ranges → two files; merge enabled → single output
- [ ] Two clips on the same URL; remove one from queue without affecting the other
- [ ] Single clip start/end still works as before
## Notes
- Requires **ffmpeg** (already in official MeTube Docker).
- Browser extension not included in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,15 @@ Key capabilities:
 * Download videos, audio, captions, and thumbnails from a browser UI.
 * Download playlists and channels, with configurable output and download options.
 * Subscribe to channels and playlists, periodically check for new items, and queue new uploads automatically.
+* Download only part of a video or audio track using **Clip start** / **Clip end** (see below).
 
 ![screenshot1](https://github.com/alexta69/metube/raw/master/screenshot.gif)
+
+## ✂️ Time clips (partial downloads)
+
+For **Video** or **Audio**, use **Clip start** / **Clip end** (`90`, `1:30`, `H:MM:SS`), or **Batch clips** in Advanced Options (several ranges for the same URL — separate files or one merged file). Prefer a URL without `&t=` when setting times manually. Not available for subscriptions.
+
+Bookmarklets or other tools can prefill the add form via `?url=…&clips=…` or `#mt=…` (see `ui/src/app/metube-import-query.ts`).
 
 ## 🐳 Run using Docker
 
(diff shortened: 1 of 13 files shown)


alexta69:
Hi @PepegaSan,
MeTube is not meant to replace video editing software -- if we start adding features like this, there's no end to it. I want to keep the program minimal and maintainable, and avoid bloat. The current start-time - end-time functionality is as far as MeTube is going to go.
Thanks, and sorry that I have to reject the PR.

Issue #222, opened 2023-02-21
Feature: Know Video Quality before download


Snake883:
Before I download a video, I would like to know which video quality (resolution/fps) is available, and what the file sizes are before I download.

I would also like to see this information in the "Downloading" and "Completed" sections.

alexta69:
I'm not sure it's generically possible with yt-dlp (esp. regarding file sizes). Leaving it open in case somebody wants to work on this (but please first update on your intention here so we can discuss how the UI is going to look like).

darmiel:
> I'm not sure it's generically possible with yt-dlp

You can use `extract_info` with the option `download` set to `False` to get a list of formats from the video, show the options to the user and then download the desired format. I've taken a rough look at the code, but I think it would require a rather large rewrite of MeTube to include this. Here's what I've tested:

```py
with yt_dlp.YoutubeDL() as ytdl:
    info = ytdl.extract_info('https://www.youtube.com/watch?v=_SxDCtMpZm0', download=False)

    # check if input was a video
    if info.get('_type', 'video') != 'video':
        raise ...

    # list formats on frontend
    # formats = '\n'.join(z['format'] for z in info['formats'])

    # A bit hacky - to avoid having to retrieve information about the video again, 
    # I overwrite the list of formats with just the requested format, then download it via the `process_ie_result` method.
    # It's probably best to just re-request the video information again and set the requested format there.
    info['formats'] = [next((f for f in info['formats'] if f['format_id'] == '251'), None)] # I wanted format 251
    assert info['formats']

    # download the video
    ytdl.process_ie_result(info)
```

**This is the easiest thing to to regarding this functionality, in my opinion:**

A route which returns a list of formats, (maybe even title and thumbnail (these are contained in the `extract_info` result anyway)), e.g. 

```http
GET /info?url=<url>

{ "title": "...", "formats": [ { ... } ] }
```

The frontend then asks for this route when inserting a URL, waits for an answer and then fills in the formats based on the previous information (and possibly even displays the title and thumbnail) - the rest should then be pretty much the same as it is now.

alexta69:
Consolidating all pre-download format/quality/track-selection requests into one tracking issue: #1032. It lays out an agreed UI direction (probe + picker modal, per-download format override) that would cover this request — contributions welcome there. Closing this one as a duplicate to keep the discussion in one place.

PR #991, opened 2026-05-29
feat(ui): show actually-downloaded resolution/bitrate in completed list

### LINKED ISSUE #745: 1080p MP4 download often falls back to 640x360
When selecting 1080p video quality and MP4 format, the download does not always retrieve the requested resolution. Most of the time, it pulls a 640x360 video instead.

To get the correct 1080p file, I have to delete the lower-resolution version and retry the download — sometimes up to 5 attempts before it finally grabs the 1080p version.

Environment:

* Running latest Docker image (2025.07.31)

* macOS host

Steps to reproduce:

1. Select a video that has 1080p available.

2. Choose 1080p quality and MP4 format.

3. Start download.

4. Check the file resolution — often it’s 640x360.

5. Repeat download until 1080p is eventually fetched.

Expected behavior:
When selecting 1080p MP4, the first download should consistently retrieve the 1080p version.

Actual behavior:
Downloads often fall back to 640x360 resolution, requiring multiple retries.

webdevsmarthome:
## Motivation

The completed-downloads list currently shows the user's *requested* quality (`Best`, `1080p`, `320 kbps`), but yt-dlp may resolve that to anything from 360p to 4K60 depending on what the source actually offers. The user has no way to tell after the fact what came down — was that "Best" video 720p30 or 2160p60?

This bites in a few common scenarios:
- `Best` selected — what did you actually get?
- `1080p` selected but only `720p` was available → silent downgrade (cf. closed #745)
- `320 kbps` audio — did you get a 320 source or did yt-dlp fall back to 256?

The information is right there in `info_dict` at the MoveFiles postprocessor hook — MeTube just doesn't persist it. This PR plumbs four optional fields through and renders the actual value next to the requested one when they differ.

### Backend (`app/ytdl.py`)

- `DownloadInfo.__init__`: five new optional fields, all defaulting to `None`:
  `width`, `height`, `fps`, `vcodec_actual`, `abr`.
- `DownloadInfo.__setstate__`: backfills `None` for items persisted before this patch so old `completed.json` entries keep loading cleanly.
- `put_status_postprocessor` (`MoveFiles` / `finished`): pulls `width`/`height`/`fps`/`abr` straight from `d['info_dict']` and adds `vcodec_actual` only when `vcodec` is present and != `'none'` (skipping audio-only formats where `vcodec='none'` just means there isn't one).
- `update_status`: applies the five fields to `self.info` without overwriting present values with `None`.

### Frontend (`ui/src/app/interfaces/download.ts`)

Five new optional fields on `Download` mirroring the backend.

### Frontend (`ui/src/app/app.ts:formatQualityLabel`)

Refactored into three helpers:
- `formatQualityLabel` — orchestrates and joins with a middle-dot
- `formatRequestedQuality` — what the user picked (unchanged logic)
- `formatActualQuality` — what yt-dlp actually delivered

Examples of the resulting labels:

| User pick | Actual | Rendered |
|---|---|---|
| Best | 1080p60 video | `Best · 1080p60` |
| Best | 720p30 video | `Best · 720p` |
| Best | 2160p60 video | `Best · 2160p60` |
| 1080 | 720p (downgrade) | `1080p · 720p` |
| 1080 | 1080p (exact) | `1080p` *(collapsed — no noise)* |
| Best audio | 320 kbps | `Best · 320 kbps` |
| Old persisted entry | — (no data) | `Best` *(unchanged)* |

`fps` is appended only at 50+ (so 60fps gets marked, 24/25/30 stay implicit — they would just clutter without telling the user anything new).

## Backward compatibility

- Old persisted `completed.json` entries don't have the new fields. `__setstate__` backfills them to `None`, and `formatActualQuality()` returns `null` in that case → label rendering is identical to before. No data migration needed.
- For chapters / captions / thumbnails the label code path was unchanged (`return '-'` short-circuits early).

## Testing

Verified locally:
- ✅ `docker compose build` succeeds, `pnpm build` clean.
- ✅ Bundle contains the new helper symbols and the static labels.
- ✅ A pre-patch completed item ("Me at the zoo", 360p) loads with just `Best` (no width/height in the persisted info → label unchanged).
- ✅ A post-patch completed item populates the new fields and renders `Best · …` as expected.
- ✅ Backend doesn't crash if `info_dict` doesn't have any of the fields (each is `info.get(...)`-guarded individually).

## Refs

- #222 "Know Video Quality before download" — related, but about *pre*-download visibility. This PR addresses the *post*-download side.
- #745 "1080p MP4 download often falls back to 640x360" — exactly the silent-downgrade case this PR makes visible.

### DIFF
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -231,10 +231,24 @@ class DownloadInfo:
         self.live_status = live_status
         self.live_release_timestamp = live_release_timestamp
         self.subtitle_files = []
+        # Actual media properties captured from yt-dlp's info_dict when the
+        # download finishes. Unlike `quality`/`codec` (the user's request),
+        # these reflect what yt-dlp actually resolved to. Set lazily by the
+        # MoveFiles postprocessor hook, so they stay None for in-flight,
+        # failed, or non-media (captions/thumbnail) downloads.
+        self.width = None
+        self.height = None
+        self.fps = None
+        self.vcodec_actual = None
+        self.abr = None
 
     def __setstate__(self, state):
         """BACKWARD COMPATIBILITY: migrate old DownloadInfo from persistent queue files."""
         self.__dict__.update(state)
+        # Backfill resolved-media fields for items persisted before this patch.
+        for field in ('width', 'height', 'fps', 'vcodec_actual', 'abr'):
+            if not hasattr(self, field):
+                setattr(self, field, None)
         if 'download_type' not in state:
             old_format = state.get('format', 'any')
             old_video_codec = state.get('video_codec', 'auto')
@@ -454,7 +468,21 @@ class Download:
                         filename = os.path.join(finaldir, os.path.basename(filepath))
                     else:
                         filename = filepath
-                    self.status_queue.put({'status': 'finished', 'filename': filename})
+                    status_update = {'status': 'finished', 'filename': filename}
+                    # Capture actual media properties resolved by yt-dlp so the
+                    # UI can show "Best · 1080p60" instead of just "Best".
+                    # Skip vcodec='none' (audio-only formats) — that's not the
+                    # video codec, it just means there isn't one.
+                    info = d['info_dict']
+                    for src, dst in (('width', 'width'), ('height', 'height'),
+                                     ('fps', 'fps'), ('abr', 'abr')):
+                        value = info.get(src)
+                        if value is not None:
+                            status_update[dst] = value
+                    vcodec = info.get('vcodec')
+                    if vcodec and vcodec != 'none':
+                        status_update['vcodec_actual'] = vcodec
+                    self.status_queue.put(status_update)
                     # For captions-only downloads, yt-dlp may still report a media-like
                     # filepath in MoveFiles. Capture subtitle outputs explicitly so the
                     # UI can link to real caption files.
@@ -585,6 +613,12 @@ class Download:
                 self.info.size = os.path.getsize(fileName) if os.path.exists(fileName) else None
                 if getattr(self.info, 'download_type', '') == 'thumbnail':
                     self.info.filename = re.sub(r'\.webm$', '.jpg', self.info.filename)
+            # Capture actual media properties yt-dlp resolved (width / height /
+            # fps / actual vcodec / audio bitrate). Pushed by the MoveFiles
+            # postprocessor hook; never overwrite a present value with None.
+            for field in ('width', 'height', 'fps', 'vcodec_actual', 'abr'):
+                if field in status and status[field] is not None:
+                    setattr(self.info, field, status[field])
 
             # Handle chapter files
             log.debug(f"Update status for {self.info.title}: {status}")
--- a/ui/src/app/app.ts
+++ b/ui/src/app/app.ts
@@ -862,9 +862,34 @@ export class App implements AfterViewInit, OnInit, OnDestroy {
     }
     const q = download.quality;
     if (!q) return '';
-    if (/^\d+$/.test(q) && download.download_type === 'audio') return `${q} kbps`;
+    const requested = this.formatRequestedQuality(q, download.download_type);
+    const actual = this.formatActualQuality(download);
+    // Show the actual value next to the requested one only when it differs
+    // — most useful for "Best" / "Worst" (no exact label) or when yt-dlp
+    // had to downgrade (user requested 1080p, only 720p was available).
+    if (!actual || actual === requested) return requested;
+    return `${requested} · ${actual}`;
+  }
+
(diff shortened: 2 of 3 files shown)


ui/src/app/app.ts
```diff
@@ -858,11 +858,36 @@ export class App implements AfterViewInit, OnInit, OnDestroy {
...
+  private formatRequestedQuality(q: string, dl_type: string): string {
+    if (/^\d+$/.test(q) && dl_type === 'audio') return `${q} kbps`;
     if (/^\d+$/.test(q)) return `${q}p`;
     return q.charAt(0).toUpperCase() + q.slice(1);
   }
 
+  // Returns the actually-resolved quality from yt-dlp's info_dict
+  // (height+fps for video, abr for audio). Returns null for old persisted
+  // entries that pre-date the backend patch, or for non-media types.
+  private formatActualQuality(download: Download): string | null {
```
alexta69, line 883: `formatActualQuality` only reads `height`, `fps`, and `abr`. The `width` and `vcodec_actual` fields added in this PR aren't read here (or anywhere in `ui/src`), so they end up persisted but never shown.

Could we drop `width` and `vcodec_actual` entirely? `height` already covers resolution and `abr` covers bitrate, which matches the PR's stated scope. If actual-codec display is something you'd like, it'd fit better as a follow-up that also wires `formatCodecLabel`.

ui/src/app/app.ts
```diff
@@ -858,11 +858,36 @@ export class App implements AfterViewInit, OnInit, OnDestroy {
...
+    if (download.download_type === 'audio') {
+      return download.abr ? `${Math.round(download.abr)} kbps` : null;
+    }
+    if (!download.height) return null;
+    let label = `${download.height}p`;
+    // High frame rates (50+) are worth highlighting; standard 24/25/30
+    // would clutter without telling the user anything new.
+    if (download.fps && download.fps >= 50) label += Math.round(download.fps);
```
alexta69, line 891: Minor/non-blocking: when the user requested an exact resolution and got the same height at a high frame rate (e.g. requested `1080p`, got 1080p60), the label renders `1080p · 1080p60`, repeating the base resolution. Just flagging in case you'd prefer to collapse to `1080p60` when the heights match.

ui/src/app/interfaces/download.ts
```diff
@@ -30,4 +30,14 @@ export interface Download {
   error?: string;
   deleting?: boolean;
   chapter_files?: { filename: string, size: number }[];
+  // Actual media properties resolved by yt-dlp at download time. Unlike
+  // `quality`/`codec` (the user's request), these reflect what came out.
+  // Optional because they're only populated for finished downloads of
+  // video/audio (not captions/thumbnails) and old persisted entries
+  // pre-dating the patch don't have them.
+  width?: number;
```
alexta69, line 40: Following from the `app.ts` comment: `width` and `vcodec_actual` have no consumers in the UI. Suggest removing both from the interface and keeping just `height`, `fps`, and `abr`.

app/ytdl.py
```diff
@@ -438,7 +452,21 @@ def put_status_postprocessor(d):
...
+                    # video codec, it just means there isn't one.
+                    info = d['info_dict']
+                    for src, dst in (('width', 'width'), ('height', 'height'),
+                                     ('fps', 'fps'), ('abr', 'abr')):
+                        value = info.get(src)
+                        if value is not None:
+                            status_update[dst] = value
+                    vcodec = info.get('vcodec')
+                    if vcodec and vcodec != 'none':
+                        status_update['vcodec_actual'] = vcodec
```
alexta69, line 484: Mirroring the frontend: if we drop `width` and `vcodec_actual`, this capture can shrink to just `('height', 'fps', 'abr')` and the `vcodec` lookup/guard can go away. The same trim then applies to the `DownloadInfo.__init__` defaults, the `__setstate__` backfill tuple, and the `update_status` apply loop — removing two fields across all four touch points.

alexta69 (changes requested):
Thanks for this — surfacing the actual vs. requested resolution is a genuinely useful addition, and the backward-compat handling (`__setstate__` backfill + the "never overwrite with None" guard) is careful and correct.

One thing to address before merge: two of the five new fields (`width` and `vcodec_actual`) are plumbed all the way through the backend and the `Download` interface but never actually rendered — `formatActualQuality()` only uses `height`, `fps`, and `abr`. Since the PR is scoped to resolution + bitrate, dropping those two would keep the change tight (5 fields → 3) without losing any displayed information. Details inline.

alexta69:
Closing as part of a stale-PR cleanup — the branch conflicts with master after the frontend rewrite. The feature is still welcome; feel free to resubmit rebased onto current master.

PR #1000, opened 2026-06-06
chore: improve metube maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in app/tests/conftest.py, app/tests/test_api.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


alexta69:
Thanks, but closing — the case this test covers is already handled and tested well enough, so it doesn't add enough value to merge.

PR #1003, opened 2026-06-08
Add YouTube Music artist subscription support

### LINKED ISSUE #1002: [Feature]: Support YouTube Music album/single releases in channel subscriptions
### Prerequisites

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before
- [x] I have read the [feature request guidelines](https://github.com/alexta69/metube#-submitting-feature-requests) in the README
- [x] I understand that MeTube development relies on community contributions and the maintainer is not likely to implement this feature
- [x] I have considered if this feature should be implemented in yt-dlp instead of MeTube (if so, please report to [yt-dlp repository](https://github.com/yt-dlp/yt-dlp/issues))

### Feature Description

Hi,

**Context**
I use MeTube primarily with YouTube Music and have been trying to use channel 
subscriptions to automatically download new releases. Unfortunately, YouTube Music 
behaves differently from regular YouTube in a way that breaks the current 
subscription flow.

**Problem**
When subscribing to a YouTube Music channel, MeTube correctly downloads videos and 
shorts, but misses Albums and Singles entirely. This is because YouTube Music 
exposes releases as playlist URLs rather than standard video URLs — a distinction 
the current subscription logic doesn't account for.

**Proposed solution**
I have written a personal script using `ytmusicapi` that resolves the correct 
playlist URLs for a given artist's releases. I would like to discuss whether this 
logic could be integrated directly into MeTube as a first-class feature.

The implementation would be scoped exclusively to YouTube Music channel 
subscriptions and would not alter any existing behavior for standard MeTube 
subscriptions.

Happy to share the script as a starting point or contribute directly if the 
approach seems viable. Let me know your thoughts!

### Implementation Plan

Maybe, I'd like to discuss the implementation approach first

### Additional Context

_No response_

Woodyy07:
Closes #1002 

When a music.youtube.com/channel/... URL is subscribed, use ytmusicapi to enumerate the artist's album and single releases instead of yt-dlp flat-extract, which only returns individual topic video uploads.

Each release is tracked as an OLAK5uy_ playlist ID in seen_ids, so new albums and singles are detected and queued automatically on each check.

- Add extract_ytmusic_artist_releases() using ytmusicapi.get_artist() and get_artist_albums() for full pagination
- Route YTMusic URLs in add_subscription() and _check_one_unlocked()
- Add ytmusicapi dependency
- Add 31 tests covering URL detection, extraction, and manager integration

### DIFF
--- a/app/subscriptions.py
+++ b/app/subscriptions.py
@@ -31,8 +31,11 @@ _MEDIA_HINT_FIELDS = (
     "live_status",
     "availability",
 )
-_YTMUSIC_ARTIST_URL_RE = re.compile(
+_YTMUSIC_CHANNEL_URL_RE = re.compile(
     r"https?://music\.youtube\.com/channel/([A-Za-z0-9_-]+)"
+)
+_YTMUSIC_HANDLE_URL_RE = re.compile(
+    r"https?://music\.youtube\.com/@([A-Za-z0-9._-]+)"
 )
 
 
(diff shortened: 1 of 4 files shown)


alexta69:
Hi @Woodyy07,
MeTube is a UI for yt-dlp; we inherit that project's support for various sites, and don't introduce site-specific logic, because it doesn't belong here -- that's precisely what yt-dlp is for. I suggest you submit this feature to the yt-dlp project instead, and then MeTube (and everyone else who uses yt-dlp) will inherit it automatically.
Thanks

Woodyy07:
Hi @alexta69,

I now have a better understanding of the project's philosophy — I wasn't aware it was meant to be purely a yt-dlp UI, so I thought a feature like this could fit in nicely.

I'll look into whether a proper implementation could be done at the yt-dlp level, though I doubt I have the knowledge or capacity to pull it off — so I'll probably stick to maintaining my own fork in the meantime!

Thanks for your work on MeTube, it's a great project that I use every day.

Issue #396, opened 2024-02-20
AUDIO_DOWNLOAD_DIR


noro-gamser:
Hey, 

may it is an user error, but I cant use:

AUDIO_DOWNLOAD_DIR=/music

version: "3"
services:
  metube:
    image: ghcr.io/alexta69/metube
    container_name: metube
    restart: unless-stopped
    ports:
      - 8081:8081
    volumes:
      - /downloads:/downloads
networks:
  dockge_default:
    external: true

.env
AUDIO_DOWNLOAD_DIR=/music 

I tried diffrent approches nothing worked out, 

I even tried to change the entery script 

#!/bin/sh

echo "Setting umask to ${UMASK}"
umask ${UMASK}
echo "Creating download directory (${DOWNLOAD_DIR}), audio download directory (${AUDIO_DOWNLOAD_DIR}), state directory (${STATE_DIR}), and temp dir (${TEMP_DIR})"
mkdir -p "${DOWNLOAD_DIR}" "${AUDIO_DOWNLOAD_DIR}" "${STATE_DIR}" "${TEMP_DIR}" "/music"

if [ `id -u` -eq 0 ] && [ `id -g` -eq 0 ]; then
    if [ "${UID}" -eq 0 ]; then
        echo "Warning: it is not recommended to run as root user; please check your setting of the UID environment variable"
    fi
    echo "Changing ownership of download, state, and audio download directories to ${UID}:${GID}"
    chown -R "${UID}":"${GID}" /app "${DOWNLOAD_DIR}" "${AUDIO_DOWNLOAD_DIR}" "${STATE_DIR}" "${TEMP_DIR}" "/music"
    echo "Running MeTube as user ${UID}:${GID}"
    su-exec "${UID}":"${GID}" python3 app/main.py
else
    echo "User set by Docker; running MeTube as `id -u`:`id -g`"
    python3 app/main.py
fi

Im not a dev spend few hours to understand what my issue is. I only want to separate the audios from the others :/ 

edit: i saw only now this:

metube  | ERROR: Unable to download video: [Errno 2] No such file or directory: '/downloads/Wind.webm.part'
metube  | INFO:ytdl:waiting for item to download

did it stilltries to download into the /downloads folder?

PikuZheng:
I don't fully understand with your description, but yt-dlp needs to write temporary files first, then use ffmpeg to convent it to a final file, and finally move it into download_dir.
If you do not configure temp_dir, it is equivalent to download_dir.
[...]

PR #1004, opened 2026-06-11
Create AUDIO_DOWNLOAD_DIR in Docker image

akeeton:
Setting the `AUDIO_DOWNLOAD_DIR` environment variable when using Docker Compose causes errors because the directory isn't created.  This PR adds `AUDIO_DOWNLOAD_DIR` to the directory creation commands in `docker-entrypoint.sh`. It also explicitly sets the default value for `AUDIO_DOWNLOAD_DIR` in the `Dockerfile` so that an unset `AUDIO_DOWNLOAD_DIR` doesn't cause errors in `docker-entrypoint.sh`.

This is the `compose.yaml` I used for testing:

```yaml
services:
  metube:
    build: .
    container_name: metube
    ports:
      - "8081:8081"
    environment:
      DOWNLOAD_DIR: /downloads/video
      AUDIO_DOWNLOAD_DIR: /downloads/audio
    volumes:
      - metube-downloads:/downloads

volumes:
  metube-downloads:
```

And the command:
```bash
docker compose build && docker compose up -d && docker compose logs -f
```

Logs before the PR changes:

> metube  | Setting umask to 022
> metube  | Creating download directory (/downloads/video), state directory (/downloads/.metube), and temp dir (/downloads)
> metube  | Changing ownership of download and state directories to 1000:1000
> metube  | Starting BgUtils POT Provider
> metube  | Running MeTube as user 1000:1000
> metube  | Traceback (most recent call last):
> metube  |   File "/usr/local/lib/python3.13/site-packages/aiohttp/web_urldispatcher.py", line 568, in __init__
> metube  |     directory = Path(directory).expanduser().resolve(strict=True)
> metube  |   File "/usr/local/lib/python3.13/pathlib/_local.py", line 670, in resolve
> metube  |     return self.with_segments(os.path.realpath(self, strict=strict))
> metube  |                               ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
> metube  |   File "<frozen posixpath>", line 451, in realpath
> metube  | FileNotFoundError: [Errno 2] No such file or directory: '/downloads/audio'
> metube  | 
> metube  | The above exception was the direct cause of the following exception:
> metube  | 
> metube  | Traceback (most recent call last):
> metube  |   File "/app/app/main.py", line 1135, in <module>
> metube  |     raise e
> metube  |   File "/app/app/main.py", line 1131, in <module>
> metube  |     app.add_routes(routes)
> metube  |     ~~~~~~~~~~~~~~^^^^^^^^
> metube  |   File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 389, in add_routes
> metube  |     return self.router.add_routes(routes)
> metube  |            ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
> metube  |   File "/usr/local/lib/python3.13/site-packages/aiohttp/web_urldispatcher.py", line 1287, in add_routes
> metube  |     registered_routes.extend(route_def.register(self))
> metube  |                              ~~~~~~~~~~~~~~~~~~^^^^^^
> metube  |   File "/usr/local/lib/python3.13/site-packages/aiohttp/web_routedef.py", line 98, in register
> metube  |     resource = router.add_static(self.prefix, self.path, **self.kwargs)
> metube  |   File "/usr/local/lib/python3.13/site-packages/aiohttp/web_urldispatcher.py", line 1211, in add_static
> metube  |     resource = StaticResource(
> metube  |         prefix,
> metube  |     ...<6 lines>...
> metube  |         append_version=append_version,
> metube  |     )
> metube  |   File "/usr/local/lib/python3.13/site-packages/aiohttp/web_urldispatcher.py", line 570, in __init__
> metube  |     raise ValueError(f"'{directory}' does not exist") from error
> metube  | ValueError: '/downloads/audio' does not exist
> metube exited with code 1

Logs after the PR changes:

> metube  | Setting umask to 022
> metube  | Creating download directory (/downloads/video), audio download directory (/downloads/audio), state directory (/downloads/.metube), and temp dir (/downloads)
> metube  | Changing ownership of download and state directories to 1000:1000
> metube  | Starting BgUtils POT Provider
> metube  | Running MeTube as user 1000:1000
> metube  | INFO:main:Listening on 0.0.0.0:8081
> metube  | INFO:ytdl:Initializing DownloadQueue
> [...]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,7 @@ ENV PGID=1000
 ENV UMASK=022
 
 ENV DOWNLOAD_DIR=/downloads
+ENV AUDIO_DOWNLOAD_DIR=/downloads
 ENV STATE_DIR=/downloads/.metube
 ENV TEMP_DIR=/downloads
 ENV PORT=8081
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,6 @@ ENV PGID=1000
 ENV UMASK=022
 
 ENV DOWNLOAD_DIR=/downloads
-ENV AUDIO_DOWNLOAD_DIR=/downloads
 ENV STATE_DIR=/downloads/.metube
 ENV TEMP_DIR=/downloads
 ENV PORT=8081
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -2,6 +2,7 @@
 
 PUID="${UID:-$PUID}"
 PGID="${GID:-$PGID}"
+AUDIO_DOWNLOAD_DIR="${AUDIO_DOWNLOAD_DIR:-$DOWNLOAD_DIR}"
 
 echo "Setting umask to ${UMASK}"
 umask ${UMASK}
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -5,8 +5,8 @@ PGID="${GID:-$PGID}"
 
 echo "Setting umask to ${UMASK}"
 umask ${UMASK}
-echo "Creating download directory (${DOWNLOAD_DIR}), state directory (${STATE_DIR}), and temp dir (${TEMP_DIR})"
-mkdir -p "${DOWNLOAD_DIR}" "${STATE_DIR}" "${TEMP_DIR}"
+echo "Creating download directory (${DOWNLOAD_DIR}), audio download directory (${AUDIO_DOWNLOAD_DIR}), state directory (${STATE_DIR}), and temp dir (${TEMP_DIR})"
+mkdir -p "${DOWNLOAD_DIR}" "${AUDIO_DOWNLOAD_DIR}" "${STATE_DIR}" "${TEMP_DIR}"
 
 do_upgrade() {
     echo "Upgrading yt-dlp to nightly channel..."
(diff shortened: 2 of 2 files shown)


alexta69 (commented):
Thanks for tracking this down — the bug is real. `routes.static()` in `app/main.py` resolves the directory with `strict=True` at startup and raises `ValueError` if it doesn't exist, and the entrypoint previously only created `DOWNLOAD_DIR`/`STATE_DIR`/`TEMP_DIR`. So pointing `AUDIO_DOWNLOAD_DIR` at a distinct, not-yet-existing path crashes the container on boot. The `mkdir -p` / `chown -R` additions are exactly the right fix.

One concern with the `Dockerfile` change (`ENV AUDIO_DOWNLOAD_DIR=/downloads`): it introduces a subtle behavior regression.

`AUDIO_DOWNLOAD_DIR` is documented as *"Defaults to the value of `DOWNLOAD_DIR`"*, and that default is implemented in Python via the `%%DOWNLOAD_DIR` sentinel (`Config._DEFAULTS`), which only resolves when the env var is **unset**. Hardcoding `AUDIO_DOWNLOAD_DIR=/downloads` in the image means the var is now always set, so the fallback never runs.

Concretely: a user who sets only `DOWNLOAD_DIR=/downloads/video` (no `AUDIO_DOWNLOAD_DIR`) currently gets audio in `/downloads/video`. After this change they'd get audio in `/downloads` instead. The plain default case is unaffected since both values coincide, but it's still a real change that contradicts the README.

The `Dockerfile` ENV was only needed to keep `mkdir`/`chown` from getting an empty quoted argument when the var is unset. A cleaner approach is to drop the Dockerfile change and derive the default in the entrypoint instead, mirroring the existing `PUID="${UID:-$PUID}"` idiom:

```sh
AUDIO_DOWNLOAD_DIR="${AUDIO_DOWNLOAD_DIR:-$DOWNLOAD_DIR}"
```

placed near the top of `docker-entrypoint.sh`. That avoids the empty-argument crash while preserving the documented "defaults to `DOWNLOAD_DIR`" behavior even when `DOWNLOAD_DIR` is customized. The `mkdir`/`chown` additions can stay as-is.

akeeton:
Thanks for taking the time to write a thorough review.  I've pushed the suggested changes.

Edit: This also probably addresses issue #396, although it's already closed.

alexta69:
Thanks @akeeton .

PR #1010, opened 2026-06-16
fix: blank PUBLIC_HOST_AUDIO_URL no longer breaks audio download links

pacnpal:
## The bug

When `PUBLIC_HOST_AUDIO_URL` is set but blank (for example `PUBLIC_HOST_AUDIO_URL=` left empty in a compose file), audio download links come out root-relative, like `app.url/song.mp3`, and 404. Video links keep working because they use `PUBLIC_HOST_URL`. So you end up with video downloads giving a working link and audio downloads giving a broken one.

## Why it happens

`os.environ.get(key, default)` only falls back to the default when the key is absent. A present-but-blank value stays `''`, and the trailing-slash normalizer skips empty strings, so `''` gets shipped to the browser. `buildDownloadLink` then uses that empty base with no fallback, and the link resolves against the page root.

## The fix

Backend: a blank `PUBLIC_HOST_AUDIO_URL` now inherits `PUBLIC_HOST_URL`, which is what the README already says it does ("Same as PUBLIC_HOST_URL but for audio downloads"). When both are blank it stays blank, so serving from the web root still works and the existing behavior is preserved.

Frontend: `buildDownloadLink` and `buildChapterDownloadLink` fall back to `PUBLIC_HOST_URL` when the audio host is blank, so an empty value can't produce a root-relative link even from the UI side.

## Known limit

If you run a separate `AUDIO_DOWNLOAD_DIR` and leave the audio host blank, inheriting `PUBLIC_HOST_URL` points audio at the video directory. That combination was already broken before this change, just root-relative instead. The answer there is to set `PUBLIC_HOST_AUDIO_URL` explicitly. I can add a directory-aware fallback if you'd prefer to cover it here.

## Tests

Added backend and UI regression tests, both confirmed to fail against the old code before the fix lands. Full backend suite passes (183), full UI suite passes (43), lint clean.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -117,6 +117,10 @@ class Config:
             if val and not val.endswith('/'):
                 setattr(self, attr, val + '/')
 
+        # A blank PUBLIC_HOST_AUDIO_URL inherits PUBLIC_HOST_URL so audio links aren't left root-relative; both blank stays blank.
+        if not self.PUBLIC_HOST_AUDIO_URL:
+            self.PUBLIC_HOST_AUDIO_URL = self.PUBLIC_HOST_URL
+
         # Convert relative addresses to absolute addresses to prevent the failure of file address comparison
         if self.YTDL_OPTIONS_FILE and self.YTDL_OPTIONS_FILE.startswith('.'):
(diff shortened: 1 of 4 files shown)


alexta69:
Thanks for highlighting the problem. I pushed another (smaller) fix.

PR #1023, opened 2026-07-08
Don't mark a subscription as broken just because all entries are filtered out as they have already been downloaded

lanthaler:
This happens e.g. if `download_archive` is used and SUBSCRIPTION_SCAN_PLAYLIST_END is too small and new entries are added to the end instead of the beginning of the playlist leading to a super confusing error message at the moment (_"This URL points to a single video, not a channel or playlist. Use Download instead."_).


alexta69:
Thanks @lanthaler !

PR #1025, opened 2026-07-15
feat: add AlbumArtistPostProcessor to fill missing album-artist metadata

jahruz67:
<html>
<body>
<h2>Motivation</h2><p>YouTube Music downloads already receive track and album metadata from yt-dlp, but the Album Artist tag is often left blank. This causes music libraries and media players to organize albums incorrectly even though the main artist is available in the extracted metadata.</p><p>This is especially noticeable in a few common scenarios:</p><ul><li>A normal album such as <em>CrasH Talk</em> has “ScHoolboy Q” as Artist but no Album Artist.</li><li>Tracks featuring guest performers may be split into separate album entries if every credited performer is treated as the Album Artist.</li><li>Compilation albums explicitly identified as “Various Artists” must retain that value.</li></ul><p>The necessary artist and album information is already present in yt-dlp’s <code>info_dict</code>; MeTube just needs to populate the missing Album Artist field before metadata is embedded.</p><h2>What changes</h2><h3>Backend (<code>app/ytdl.py</code>)</h3><p>Adds a private yt-dlp pre-processing postprocessor that runs for audio downloads before FFmpeg metadata embedding.</p><p>When a track has an album but no existing <code>album_artist</code> or <code>album_artists</code> value, the postprocessor:</p><ul><li>Reads yt-dlp’s <code>artists</code> or <code>artist</code> metadata.</li><li>Selects the leading/main credited artist.</li><li>Handles YouTube Music’s middle-dot-separated credits.</li><li>Writes the result to <code>album_artist</code>.</li></ul><p>Existing Album Artist metadata is never replaced, including an explicit “Various Artists” value.</p><p>The fallback does not use uploader or channel names, preventing values such as <code>ScHoolboy Q - Topic</code> from being written as the Album Artist.</p><h3>Examples</h3><div><div>
Album | Track artists | Album Artist
-- | -- | --
CrasH Talk | ScHoolboy Q | ScHoolboy Q
CrasH Talk | ScHoolboy Q · Travis Scott | ScHoolboy Q
Compilation | Various Artists | Various Artists
No album metadata | Standalone Artist | Unchanged

</div></div><h2>Backward compatibility</h2><p>No persistence schema, API, configuration, or frontend changes are required.</p><p>Previously downloaded files are not retagged. The new behavior applies only to future audio downloads processed through yt-dlp’s metadata pipeline.</p><p>Downloads without album metadata are unchanged. Video, caption, and thumbnail downloads are also unaffected.</p>
</body>
</html>

### DIFF
--- a/app/tests/test_ytdl_utils.py
+++ b/app/tests/test_ytdl_utils.py
@@ -87,6 +87,40 @@ class AlbumArtistPostProcessorTests(unittest.TestCase):
 
         self.assertEqual(result['album_artist'], 'ScHoolboy Q')
 
+    def test_uses_topic_channel_artist_for_joint_album(self):
+        info = {
+            'album': 'Watch the Throne',
+            'artists': ['JAY-Z', 'Kanye West'],
+            'channel': 'JAY-Z & Kanye West - Topic',
+        }
+
+        _, result = self.postprocessor.run(info)
+
+        self.assertEqual(result['album_artist'], 'JAY-Z & Kanye West')
+
+    def test_uses_topic_uploader_and_strips_suffix_for_compilation(self):
+        info = {
+            'album': 'Compilation',
+            'artist': 'Track Artist',
+            'channel': 'Regular Channel',
+            'uploader': 'Various Artists - Topic',
+        }
+
+        _, result = self.postprocessor.run(info)
+
+        self.assertEqual(result['album_artist'], 'Various Artists')
+
+    def test_regular_channel_falls_back_to_main_artist(self):
+        info = {
+            'album': 'Album',
+            'artist': 'Track Artist',
+            'channel': 'Label Channel',
+        }
+
+        _, result = self.postprocessor.run(info)
+
+        self.assertEqual(result['album_artist'], 'Track Artist')
+
     def test_preserves_explicit_various_artists(self):
         info = {
             'album': 'Revenge of the Dreamers III',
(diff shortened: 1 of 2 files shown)


alexta69:
Thanks for the PR — the code is clean, well-tested, and correctly integrated with the FFmpegMetadata pipeline. I'd like one change to the heuristic before merging, because there's a signal available that gets more cases right.

"First credited artist = album artist" misbehaves in two scenarios:

1. **Joint albums.** A collaboration album carries both artists in `artists` on every track, so only the first gets written as Album Artist.
2. **Compilations without an explicit "Various Artists" tag.** Each track gets its own artist as Album Artist, so the album still splits per artist in the library.

Both are fixable, because for auto-generated YouTube music the `channel`/`uploader` field is `"<Album Artist> - Topic"` — YouTube's own album-level artist assignment. Featured tracks stay on the main artist's topic channel, and compilations are on "Various Artists - Topic", so it's right in all the cases above. And since YouTube only provides `album` metadata on auto-generated uploads anyway, it covers essentially every case this postprocessor triggers on. I saw you excluded uploader/channel to avoid writing "ScHoolboy Q - Topic" verbatim — right worry, but used conditionally it's exactly the field we want.

Concretely:

1. Keep the existing guards (album present, no existing `album_artist`/`album_artists`).
2. If `channel` (or `uploader`) ends with `" - Topic"`, strip the suffix and use that. The suffix check scopes this to auto-generated music and makes it a no-op for regular channels and other extractors.
3. Otherwise fall back to the first credited artist, as the PR does now — this keeps every album-tagged track tagged on non-YouTube extractors too.

That way every audio download with an album gets an Album Artist, and the topic-channel rule fixes the joint-album and compilation cases where the artist-list guess goes wrong. Happy to merge with that change plus tests for the topic-channel and suffix-stripping cases.

jahruz67:
Thanks for the feedback. I've just finished addressing it.

alexta69:
Thanks @jahruz67 !

Issue #1026, opened 2026-07-16
Music metadata lookup and tagging for audio downloads


tjelite1986:
## Feature proposal

When MeTube is used as a music grabber (audio downloads into `AUDIO_DOWNLOAD_DIR`), the resulting files usually carry the video title as their only metadata, which makes them messy in music players and libraries.

I'd like to contribute a music-tagging feature for finished audio downloads:

- A music-note action on finished audio rows opens a **Music tags** dialog.
- The dialog shows the **source description** and pre-fills the fields with any song metadata the site provided (`artist`/`track`/`album`/`release_year`/`genre` from yt-dlp).
- A search button queries **iTunes Search + Deezer** (both keyless public APIs, queried only on user action) and lists candidates (cover, title, artists, album, year, genres) to fill the fields with one click.
- Apply rewrites the tags **in place with mutagen** (already a dependency), clearing the video-site leftovers (description/comment/purl) that `--embed-metadata` writes.
- Optionally the file is moved into an `Artist/Album(Year)/Artist - Album(Year) - Title.ext` layout inside its download directory, with the result link updated.

Backend cost is small: one module (`music_meta.py`), three routes, and one method on `DownloadQueue`. To keep the source data available after a restart, the completed queue persists the compact entry (music fields plus the description truncated to 2 kB) — measured at well under 3 kB per completed item.

I have a working implementation with tests (backend + frontend, all existing suites pass), verified end-to-end in Docker. PR incoming — happy to adjust scope or drop the organize option if you prefer a leaner version.

alexta69:
Declining this — see my comment on #1027 for the full reasoning. Short version: improving metadata *at download time* is in scope (#1025 is the model), but post-download tag editing, external metadata lookups, and library organization are deliberately out — that's beets/Picard/Lidarr territory, and they do it better than a downloader UI can. Closing as not planned, with genuine thanks for the proposal and the quality of the implementation.

tjelite1986:
:) okey :)

tjelite1986:
> Declining this — see my comment on [#1027](https://github.com/alexta69/metube/pull/1027) for the full reasoning. Short version: improving metadata _at download time_ is in scope ([#1025](https://github.com/alexta69/metube/pull/1025) is the model), but post-download tag editing, external metadata lookups, and library organization are deliberately out — that's beets/Picard/Lidarr territory, and they do it better than a downloader UI can. Closing as not planned, with genuine thanks for the proposal and the quality of the implementation.

[...]

PR #1027, opened 2026-07-16
feat: music metadata lookup and tagging for finished audio downloads

tjelite1986:
Implements #1026.

## What

Finished audio downloads get a music-note action that opens a **Music tags** dialog:

- Shows the **source description** (collapsible) and pre-fills the tag fields from the site-provided song metadata (`artist`/`track`/`album`/`release_year`/`genre`).
- **Search** queries iTunes Search + Deezer (keyless, in parallel, 8 s timeout, only on user action) and lists up to 10 deduplicated candidates with cover art; clicking one fills the fields.
- **Apply tags** rewrites the file in place with mutagen (existing dependency) — multi-value artist/genre tags, `albumartist`, and clearing of the video-site leftovers (`description`/`comment`/`purl`) that `--embed-metadata` writes.
- Optional **Move into Artist/Album folders**: files the track as `Artist/Album(Year)/Artist - Album(Year) - Title.ext` inside its download directory (path-safety checked with the existing `_is_within_directory`), updating `info.filename` so the result link keeps working.

## How

- `app/music_meta.py` (new): provider lookups, artist/list splitting, filesystem-safe path parts, mutagen tagging.
- `app/main.py`: `GET music-meta` (search), `GET music-meta/source` (description + site metadata of a done download), `POST music-tag` (apply).
- `app/ytdl.py`: `DownloadQueue.music_source()` / `music_tag()`; the completed queue now persists the compact entry (music fields + description truncated to 2 kB, < 3 kB per item) so tagging still has its source data after a restart.
- UI: dialog + row action in `app.ts`/`app.html`, service methods, `MusicCandidate`/`MusicSource` interfaces. Subscribe callbacks call `cdr.markForCheck()` per the app's OnPush convention.

## Tests

- `app/tests/test_music_meta.py` (new): splitting, sanitizing, layout, tag building, real mutagen retag of a generated MP3.
- Updated the two persistence tests for the compact-entry change.
- All suites pass: 270 backend tests, 40 frontend tests, lint clean, README stays under the 25 k Docker Hub limit (24.5 k).

## Verified end-to-end

Built the Docker image and drove the real flow with Playwright against a SoundCloud track: download as mp3 → dialog shows description + site metadata → iTunes/Deezer candidates render → apply with organize → mutagen tags verified on disk, file at `Forss/Belle et Fou (Original Soundtrack)(2007)/… - Flickermood.mp3`, result link still valid, and source data still present after a container restart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ Key capabilities:
 * Download videos, audio, captions, and thumbnails from a browser UI.
 * Download playlists and channels, with configurable output and download options.
 * Subscribe to channels and playlists, periodically check for new items, and queue new uploads automatically.
+* Tag finished audio downloads with song metadata (looked up on iTunes/Deezer) and optionally file them into Artist/Album folders.
 
 ![screenshot1](https://github.com/alexta69/metube/raw/master/screenshot.gif)
 
--- a/app/main.py
+++ b/app/main.py
@@ -21,6 +21,7 @@ from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 from watchfiles import DefaultFilter, Change, awatch
 
 import bg_tasks
+import music_meta
 from ytdl import DownloadQueueNotifier, DownloadQueue, Download
 from subscriptions import SubscriptionManager, SubscriptionNotifier, SubscriptionInfo, coerce_optional_bool
 from yt_dlp.version import __version__ as yt_dlp_version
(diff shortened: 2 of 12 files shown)


tjelite1986:
Pushed a follow-up commit: a stale-response token guard so out-of-order lookup responses are discarded (a slow earlier search can no longer overwrite a faster later one, and responses arriving after the dialog closes or switches download are dropped). Lint, build, and all 40 frontend tests still pass.

alexta69:
Thanks for this — and I mean that sincerely. The execution here is genuinely careful: the path handling goes through the existing `_is_within_directory`, the UI follows the app's OnPush conventions, the tests are real (including the mutagen retag), and the README stays under the Docker Hub limit. This is one of the better-built PRs this repo has received. I'm still going to decline it, and I want to be clear that the reason is scope, not quality.

MeTube's contract is: give it a URL, it runs yt-dlp well, and correct files appear. Features that improve *the write itself* are in scope — that's why #1025 (album-artist at download time, from data the extractor already has, no UI, no new dependencies) merged. What this PR adds is a workflow *on files after they exist*: a tag editor, an external metadata search, and a library organizer. That's the core competency of beets, MusicBrainz Picard, and Lidarr, which do it properly — acoustic fingerprinting, album-level matching, batch operations — where a per-file dialog here can't reasonably compete (a 40-track playlist means 40 modals). The line I want to hold is: MeTube writes the file once, as correctly as it can; what happens to the file afterwards belongs to dedicated tools.

There are also concrete ownership costs I don't want to take on: two keyless, unversioned third-party APIs (the code already has to work around Deezer's content-type quirk — when Apple rate-limits or Deezer changes shape, that becomes this issue tracker's problem, indefinitely), egress from every self-hosted instance to Apple/Deezer, and re-persisting entry data in the completed queue, which was deliberately removed recently. And realistically, merging this invites the follow-ups — cover art, MusicBrainz, batch tagging, folder templates — each a small step further into being a media manager. The cheapest place to stop that is before step one.

I did consider whether a slimmer cut (no lookups, no organize) could land, and decided against that too — even a minimal tag editor is on the wrong side of the same line.

Your implementation is clean enough to live as a fork or a companion container watching the download directory, and I'd happily see it exist there. For most users with this need, pointing beets or Picard at `AUDIO_DOWNLOAD_DIR` is the durable answer — I'm open to a short README section documenting that pairing, if you'd like to contribute one.

Thanks again for the evident care here — sorry to say no to work of this quality, and I hope the scope rationale reads as a decision about the project's shape rather than about your contribution.

tjelite1986:
Completely fair, and thanks for taking the time to write out the reasoning — the scope line ("MeTube writes the file once, as correctly as it can; what happens afterwards belongs to dedicated tools") is a clear and useful contract, and the ownership-cost argument about keyless third-party APIs is hard to argue with. The feature will live on in my fork, which suits my setup fine.

I'll take you up on the README offer and send a short section documenting the beets/Picard pairing against AUDIO_DOWNLOAD_DIR.

PR #1028, opened 2026-07-16
feat: first-class SponsorBlock toggle

tjelite1986:
## What

Adds a "Remove sponsor segments" switch to the download form (next to "Split by chapters"). When enabled, the download runs with the same postprocessor pair the CLI's `--sponsorblock-remove sponsor` builds: `SponsorBlock` (categories: `sponsor`, `when: after_filter`) + `ModifyChapters` (`remove_sponsor_segments`).

Today this is only reachable through custom yt-dlp option presets; a first-class toggle makes the most common case (skip sponsor reads on YouTube) a one-click option.

## Notes

- The flag is threaded as a keyword argument through `add()`/`__add_entry()` so all positional call sites stay untouched, stored on `DownloadInfo` (with a `hasattr` backfill for old persisted records, like `split_by_chapters`), persisted, and carried into retries.
- Cookie-persisted in the UI like the other form options.
- Scope: direct downloads only — subscriptions can get the same flag as a follow-up if you want it there too.
- All suites pass: 290 backend tests, 40 frontend tests, lint clean. Postprocessor dicts validated against yt-dlp's `YoutubeDL` constructor.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -973,6 +973,7 @@ async def subscribe(request):
         subtitle_mode=o['subtitle_mode'],
         ytdl_options_presets=o['ytdl_options_presets'],
         ytdl_options_overrides=o['ytdl_options_overrides'],
+        sponsorblock=o['sponsorblock'],
         title_regex=post.get('title_regex'),
         skip_subscriber_only=skip_subscriber_only,
         clip_start=sub_clip_start,
--- a/app/main.py
+++ b/app/main.py
@@ -725,6 +725,7 @@ def parse_download_options(post: dict) -> dict:
     playlist_item_limit = post.get('playlist_item_limit')
     auto_start = post.get('auto_start')
     split_by_chapters = post.get('split_by_chapters')
+    sponsorblock = bool(post.get('sponsorblock'))
     chapter_template = post.get('chapter_template')
     subtitle_language = post.get('subtitle_language')
     subtitle_mode = post.get('subtitle_mode')
@@ -845,6 +846,7 @@ def parse_download_options(post: dict) -> dict:
         'playlist_item_limit': playlist_item_limit,
         'auto_start': auto_start,
         'split_by_chapters': split_by_chapters,
+        'sponsorblock': sponsorblock,
         'chapter_template': chapter_template,
         'subtitle_language': subtitle_language,
         'subtitle_mode': subtitle_mode,
@@ -890,6 +892,7 @@ async def add(request):
         o['ytdl_options_overrides'],
         o['clip_start'],
         o['clip_end'],
+        sponsorblock=o['sponsorblock'],
     )
     return web.Response(text=serializer.encode(status))
 
--- a/app/subscriptions.py
+++ b/app/subscriptions.py
@@ -182,6 +182,7 @@ class SubscriptionInfo:
     auto_start: bool = True
     playlist_item_limit: int = 0
     split_by_chapters: bool = False
+    sponsorblock: bool = False
     chapter_template: str = ""
     subtitle_language: str = "en"
     subtitle_mode: str = "prefer_manual"
@@ -242,6 +243,7 @@ def _subscription_to_record(sub: SubscriptionInfo) -> dict[str, Any]:
         "auto_start": sub.auto_start,
         "playlist_item_limit": sub.playlist_item_limit,
         "split_by_chapters": sub.split_by_chapters,
+        "sponsorblock": sub.sponsorblock,
         "chapter_template": sub.chapter_template,
         "subtitle_language": sub.subtitle_language,
         "subtitle_mode": sub.subtitle_mode,
@@ -487,6 +489,7 @@ class SubscriptionManager:
         ytdl_options_overrides: Optional[dict[str, Any]] = None,
         clip_start: Optional[float] = None,
         clip_end: Optional[float] = None,
+        sponsorblock: bool = False,
     ) -> tuple[list[str], list[str]]:
         queued_ids: list[str] = []
         queue_errors: list[str] = []
@@ -519,6 +522,7 @@ class SubscriptionManager:
                 ytdl_options_overrides,
                 clip_start,
                 clip_end,
+                sponsorblock=sponsorblock,
             )
             if isinstance(result, dict) and result.get("status") == "error":
                 msg = str(result.get("msg") or f"Queueing failed for {vurl}")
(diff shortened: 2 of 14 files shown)


alexta69:
Thanks — this is in scope and well built. It shapes the write itself, and since it only passes postprocessor params to yt-dlp's own SponsorBlock support, the API client remains yt-dlp's to maintain, not MeTube's. The implementation follows the `split_by_chapters` pattern faithfully everywhere it needs to: options parsing, the `DownloadInfo` field with `__setstate__` backfill, the safe-deserialization list, cookie persistence, and the redownload path.

One fix before merging: **postprocessor ordering when both toggles are enabled.** The SponsorBlock pair is appended after the existing `FFmpegSplitChapters` append, and yt-dlp runs same-stage postprocessors in list order — so with "Split by chapters" and "Remove sponsor segments" both on, the split runs first and `ModifyChapters` runs second. The CLI does it the other way around (`--sponsorblock-remove` + `--split-chapters` removes segments first, then splits), and the current order means the chapter files keep sponsor content and the removal desyncs the remaining chapter timings. Please move the SponsorBlock/ModifyChapters pair above the split-chapters block, and add a test asserting the combined ordering.

Two non-blocking notes: hardcoding the `sponsor` category is the right first cut — let's not grow a category picker until users ask. And subscriptions not carrying the flag is fine for this PR — I'd welcome the follow-up you offered; subscriptions are arguably where this option is most useful, since those downloads run unattended.

Happy to merge with the ordering fix.

alexta69:
Nudging this one — apologies for the long gap on my side. The PR is still one fix away from merging, and I would rather land it than let it go stale.

To recap what is outstanding: **postprocessor ordering when both toggles are on.** The SponsorBlock pair is appended after the existing `FFmpegSplitChapters`, so with "Remove sponsor segments" and "Split by chapters" both enabled, the split runs against the unmodified chapter list and the resulting files still contain the sponsor segments. `SponsorBlock` and `ModifyChapters` need to sit above the split, matching what the CLI builds for `--sponsorblock-remove sponsor --split-chapters`. A test covering the combined ordering would go with it.

If you would rather not carry it further, say so and I will finish that part myself with you credited as the author — the rest of the PR is good work and I do not want it lost.

tjelite1986:
Rebased on master and fixed the ordering. The SponsorBlock/ModifyChapters pair now sits above the chapter-splitting block, so `ModifyChapters` rewrites the chapter list before `FFmpegSplitChapters` cuts the file up — the order the CLI builds for `--sponsorblock-remove sponsor --split-chapters`. Added `SponsorBlockPostprocessorTests` covering the disabled case, the exact postprocessor pair, and the combined ordering; the ordering test fails against the old arrangement.

The rebase also picked up your move of retry to the backend, so the flag is now carried by `DownloadQueue.retry()` instead of the frontend payload, with a test in `test_download_queue.py`. 389 backend tests, 44 frontend tests, lint clean.

Happy to do the subscriptions follow-up as a separate PR once this lands.

alexta69:
The ordering fix is right, and I checked it rather than taking it on faith: the pair now sits where the CLI puts it for `--sponsorblock-remove sponsor --split-chapters` (`SponsorBlock` at `after_filter`, then `ModifyChapters` and `FFmpegSplitChapters` both in `post_process`, in that order), and your new ordering test does fail against the old arrangement — I put the bug back to confirm it. The rebase onto the retry change looks clean too. 389 backend, 44 frontend, lint clean here.

One thing I want folded in before this lands, and it's a change of position on my part — I said subscriptions could be a follow-up, and I was wrong to.

The toggle sits in the advanced-options panel, and that panel is shared: **Subscribe submits the same form.** `/subscribe` doesn't accept `sponsorblock`, so the switch is visible while subscribing and silently does nothing. That's exactly the trap I just spent #1049 removing — clip bounds sat visible in that same panel doing nothing for subscriptions, and it was my own carve-out that put them there. I don't want to add a second instance of it in the same release I removed the first.

Since you offered the follow-up anyway, and subscriptions are where sponsor removal is most useful — those downloads run unattended — could you carry it here? aac9c63 is the shape to copy:

- a `sponsorblock` field on `SubscriptionInfo`, threaded through the check flow to `add_entry`
- `/subscribe` reading it from `parse_download_options` and passing it to `add_subscription`
- `sponsorblock` in the body built by `subscriptions.service.ts`
- stored subscriptions predating the field taking the default, which `_from_stored` gives you by filtering on field name

Happy to merge as soon as that's in. If you'd rather not, say so and I'll add it myself with your authorship kept on the rest.

tjelite1986:
Carried it here — no need to add it yourself. Rebased onto master first, so this now sits on top of today's four merges; subscriptions support is the second commit.

Copied aac9c63's shape:

- `sponsorblock: bool = False` on `SubscriptionInfo`, in the same defaults block as the clip bounds — `_from_stored` filters by field name, so subscriptions stored before this take the default and nothing needs migrating
- persisted through `_subscription_to_record`
- `/subscribe` reads it from `parse_download_options` and hands it to `add_subscription`; from there it rides the check flow through `_queue_entries` into `add_entry` for every entry a check queues
- `sponsorblock` in the body `subscriptions.service.ts` builds — `SubscribePayload` extends `AddDownloadPayload`, so the field arrives typed without a second declaration
- the update endpoint's field list is unchanged, same as the clip bounds: it is set at creation time

Tests for it:

- `test_api.py` — `/subscribe` passes the flag through, and defaults it off when the form omits it
- `test_subscriptions.py` — a check queues entries with the subscription's flag, and the flag survives a reload while an older record without the field still loads as `False`
- `subscriptions.service.spec.ts` — the request body carries the flag, and carries `false` by default

Green here: 417 backend, 58 frontend, lint clean.

alexta69:
Merged locally — thank you for carrying the subscriptions half rather than handing it back.

Reviewed the whole thing again after the rebase. The subscribe path is complete now: the field is on `SubscriptionInfo` with the same defaults-block treatment as the clip bounds, `/subscribe` accepts it, `subscriptions.service.ts` sends it, and `_check_one_inner` carries it onto every queued entry — so the control in the shared panel now means the same thing on both forms, which was the whole point of sending it back.

Two things I checked rather than took on trust:

- The postprocessor ordering test is real, not incidental. I moved the SponsorBlock block below the chapter-splitting block and `test_segment_removal_runs_before_the_chapter_split` failed, which is what I wanted to see.
- Full suites on the merge: 427 backend, 58 frontend, lint clean, production build clean.

Goes out with the next release.

PR #1031, opened 2026-07-17
feat: add music metadata processing and writing for YouTube Music albums

jahruz67:
This PR introduces a new `music_metadata` module that enriches audio downloads with supplemental music metadata (structured artists, track numbers, album info) sourced entirely from information already provided by yt-dlp or MeTube's queued playlist entries — no external lookups are performed.

## What's new

- **`app/music_metadata.py`** — New module providing:
  - `MusicMetadataPreProcessor` — Enriches the yt-dlp info-dict before download with:
    - Album detection from YouTube Music's OLAK5uy_ playlist IDs and MPRE browse paths
    - Track number resolution (inline n/total, playlist index, or explicit track_number)
    - Structured artist extraction (splits YouTube Music's spaced middle-dot formatting, preserves unspaced names like half·alive)
    - Square thumbnail preference for music content (largest square moved to preferred slot)
    - Album title fallback from playlist title for confirmed albums
  - `MusicMetadataWriterPostProcessor` — Writes supplemental tags after yt-dlp moves the completed audio file, supporting MP3 (ID3 TXXX:Artists + TRCK), M4A (iTunes ----:ARTISTS + trkn), FLAC/Opus (Vorbis ARTISTS/TRACKNUMBER/TRACKTOTAL)
  - Conservative approach: only writes tags when structured artists or track numbers are present, never modifies display artist/album artist/artwork

- **`app/ytdl.py`** — Registers both new post-processors for audio downloads (pre-process + after_move)

- **`app/tests/test_music_metadata.py`** — 306 lines of tests covering:
  - Album detection (MPRE, OLAK, false positives for regular playlists)
  - Track number resolution (official wins over album order, inline totals preserved)
  - YouTube Music playlist vs. album distinction
  - Regular YouTube content remains unchanged
  - Thumbnail selection (prefers largest square)
  - Structured artist parsing (spaced middle-dot only)
  - MP3 round-trip preserving existing TPE1/TPE2/APIC
  - M4A mapping preserving existing artist/album artist/artwork
  - FLAC/Opus Vorbis mapping preserving existing fields
  - Writer dispatch (empty/unsupported skipped, enriched dispatched)

- **`app/tests/test_ytdl_utils.py`** — Updated to verify both new post-processors are registered in the correct order with the correct when values

## Design principles

- **Conservative**: Only consumes metadata yt-dlp already supplies; no external APIs, no guesswork
- **Non-destructive**: Preserves existing display artist, album artist, and artwork in all tag formats
- **Format-aware**: Targeted tag writing per format (ID3, MP4 atoms, Vorbis comments) using Mutagen
- **YouTube Music aware**: Recognizes album signals (OLAK5uy_ playlist prefix, MPRE browse paths) to apply album-specific logic"

### DIFF
--- a/app/music_metadata.py
+++ b/app/music_metadata.py
@@ -1,0 +1,275 @@
+"""Conservative music metadata enrichment for audio downloads.
+
+This module only consumes metadata already supplied by yt-dlp or retained on
+MeTube's queued playlist entry. It intentionally performs no external lookup.
+"""
+
+from __future__ import annotations
+
+import os
+from typing import Any, Optional
+from urllib.parse import urlparse
+
+from mutagen import MutagenError
+from mutagen.flac import FLAC
+from mutagen.id3 import ID3, ID3NoHeaderError, TRCK, TXXX
+from mutagen.mp4 import AtomDataType, MP4, MP4FreeForm
+from mutagen.oggopus import OggOpus
+from yt_dlp.postprocessor.common import PostProcessor
+from yt_dlp.utils import PostProcessingError
+
+
+_ARTISTS_KEY = '__metube_track_artists'
+_TRACK_NUMBER_KEY = '__metube_track_number'
+_TRACK_TOTAL_KEY = '__metube_track_total'
+_YOUTUBE_MUSIC_HOSTS = frozenset(('music.youtube.com', 'music.youtube-nocookie.com'))
+
+
+def _has_value(value: Any) -> bool:
+    if isinstance(value, str):
+        return bool(value.strip())
+    if isinstance(value, (list, tuple)):
+        return any(_has_value(item) for item in value)
+    return value is not None
+
+
+def _positive_int(value: Any) -> Optional[int]:
+    if isinstance(value, bool):
+        return None
+    try:
+        number = int(value)
+    except (TypeError, ValueError):
+        return None
+    return number if number > 0 else None
+
+
+def _track_position(value: Any) -> tuple[Optional[int], Optional[int]]:
+    """Return a track number and optional total from a scalar or ``n/total``."""
+    if isinstance(value, str) and '/' in value:
+        number, total = value.split('/', 1)
+        return _positive_int(number.strip()), _positive_int(total.strip())
+    return _positive_int(value), None
+
+
+def _is_youtube_music_url(value: Any) -> bool:
+    if not isinstance(value, str) or not value.strip():
+        return False
+    try:
+        return (urlparse(value).hostname or '').lower() in _YOUTUBE_MUSIC_HOSTS
+    except ValueError:
+        return False
+
+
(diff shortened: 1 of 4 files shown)


alexta69:
I appreciate how deliberately this was scoped after the #1027 discussion — download-time only, no external lookups, no UI. It satisfies the letter of the scope line, so this isn't a scope rejection. But I'm not going to merge it as-is; I'd take a substantially leaner version. Three reasons:

1. **The album detection is extractor work that belongs upstream.** Recognizing YouTube Music albums by the `OLAK5uy_` playlist-ID prefix and `/browse/MPRE` paths means parsing YouTube's internal identifiers — exactly what yt-dlp's extractors exist to do, and exactly what breaks silently when YouTube shuffles its scheme, leaving this repo to maintain a shadow extractor. If yt-dlp doesn't surface track numbers for YT Music album playlists, the durable fix is a yt-dlp contribution that helps every downstream tool at once. I'd genuinely encourage you to submit that upstream.

2. **The mutagen writer re-introduces per-format tag-writing that MeTube shouldn't own.** yt-dlp's own `FFmpegMetadata` already embeds `track` from `track_number` — so the pre-processor alone, enriching the info dict, gets track numbers (including `n/total`) into files through the existing pipeline with zero new writing code. What the `after_move` writer uniquely adds is the multi-value `ARTISTS` tag — a convention for Picard/beets users, who by definition already run a real tagger (see the README section on pairing MeTube with one).

3. **A failure-mode bug:** the writer raises `PostProcessingError` on a tag-write failure, which fails the entire download after the file downloaded successfully. Supplemental processing must never do that — warn and continue. (This concern goes away if the writer is dropped per point 2.)

So the version I'd merge is: `MusicMetadataPreProcessor` only — track-number/total resolution, the album-title fallback, and the square-thumbnail preference (cheap, uses only existing data, improves embedded art) — flowing through yt-dlp's existing embedding, with the mutagen writer and its tests dropped. That's most of the user-visible value at a fraction of the surface. I've also just encoded the scope principles this reflects in AGENTS.md, so agent-assisted contributions can check against them up front.

Thanks for reading the room on #1027 and adjusting — that's exactly how I hoped that discussion would land.

jahruz67:
I made the code more like you asked :)

alexta69:
Merged — thank you. This is precisely the lean version I described: the pre-processor enriches the info dict (track number/total, album fallback, square-thumbnail preference) and rides yt-dlp's existing `FFmpegMetadata` embedding, with the mutagen writer and its download-failing `PostProcessingError` gone. No new dependency, and the conservative gating — only acting when there's a real extractor album signal, so ordinary playlists aren't relabelled as albums — is exactly the right instinct. The tests read well. Thanks again for taking the #1027 discussion in stride and reshaping this; it landed exactly where I'd hoped.

Issue #980, opened 2026-05-10
[Bug]: Periodical hiccup


janusn:
### Prerequisites

- [x] I have searched existing issues and discussions to ensure this bug hasn't been reported before
- [x] I have read the [troubleshooting section](https://github.com/alexta69/metube#-troubleshooting-and-submitting-issues) in the README
- [x] I have tested this issue with yt-dlp directly (not just through MeTube UI) as described in the README
- [x] I have checked that this is not a yt-dlp issue (if it is, please report it to [yt-dlp repository](https://github.com/yt-dlp/yt-dlp/issues) instead)

### Bug Description

From my experience, the WebUI locks up periodically from time to time. Furthermore, I have been using [Home Assistant](https://hub.docker.com/r/homeassistant/home-assistant) for health check which curl the webui of metube periodically to verify how frequently the hiccup happens. In fact, I have 2 instances of Home Assistants and 2 instances of MeTube's running at `http://10.27.0.61:8081` and `http://10.27.0.61:8082`.

Here is a relevant snip of the log of an instant of home assistant:
```
2026-05-10 11:21:44.388 WARNING (MainThread) [homeassistant.helpers.entity] Update of binary_sensor.x11dpu_metube is taking over 10 seconds
2026-05-10 11:21:44.388 WARNING (MainThread) [homeassistant.helpers.entity] Update of binary_sensor.x11dpu_metube_proxy is taking over 10 seconds
2026-05-10 11:21:45.290 ERROR (MainThread) [homeassistant.components.rest.data] Timeout while fetching data: http://10.27.0.61:8081
2026-05-10 11:21:45.291 ERROR (MainThread) [homeassistant.components.rest.data] Timeout while fetching data: http://10.27.0.61:8082
2026-05-10 11:22:14.389 WARNING (MainThread) [homeassistant.helpers.entity] Update of binary_sensor.x11dpu_metube is taking over 10 seconds
2026-05-10 11:22:14.391 WARNING (MainThread) [homeassistant.helpers.entity] Update of binary_sensor.x11dpu_metube_proxy is taking over 10 seconds
2026-05-10 11:22:15.290 ERROR (MainThread) [homeassistant.components.rest.data] Timeout while fetching data: http://10.27.0.61:8082
2026-05-10 11:22:15.290 ERROR (MainThread) [homeassistant.components.rest.data] Timeout while fetching data: http://10.27.0.61:8081
[...]

PR #1035, opened 2026-07-19
fix: release per-download status_queue proxy on close to stop file-descriptor leak

### LINKED ISSUE #485: too many open files error, makes metube selfterminate
If you download a very long playlist, in my case the DUST sci-fi archives meTube v1.240620 running on QNAP x64 will eventually choke on itself, thus not be able to handle its' own database calls and exit: 
<code>
         n = pdAdSu3kY1sHV-U8 ; player = https://www.youtube.com/s/player/d2e656ee/player_ias.vflset/en_US/base.js
WARNING: [youtube] CeloPsVCZ3A: nsig extraction failed: You may experience throttling for some formats
         n = hQ7WQeIieuhkyUsa ; player = https://www.youtube.com/s/player/d2e656ee/player_ias.vflset/en_US/base.js
WARNING: [youtube] Failed to download m3u8 information: <urllib3.connection.HTTPSConnection object at 0x7fc3c67a4ed0>: Failed to establish a new connection: [Errno 24] Too many open files
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-3' coro=<DownloadQueue.__import_queue() done, defined at /share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py:213> exception=error("db file doesn't exist; use 'c' or 'n' flag to create a new db")>
Traceback (most recent call last):
  File "/share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py", line 215, in __import_queue
  File "/share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py", line 317, in add
  File "/share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py", line 295, in __add_entry
  File "/share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py", line 189, in put
  File "/opt/Apache83/lib/python3.11/shelve.py", line 243, in open
  File "/opt/Apache83/lib/python3.11/shelve.py", line 227, in __init__
  File "/opt/Apache83/lib/python3.11/dbm/__init__.py", line 85, in open
dbm.error: db file doesn't exist; use 'c' or 'n' flag to create a new db
INFO:ytdl:downloading Sci-Fi Short Film “Sleepworking” | DUST
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-2' coro=<DownloadQueue.__download() done, defined at /share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py:366> exception=OSError(24, 'Too many open files')>
Traceback (most recent call last):
  File "/share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py", line 374, in __download
  File "/share/CACHEDEV3_DATA/.qpkg/meTube83/metube/app/ytdl.py", line 108, in start
  File "/opt/Apache83/lib/python3.11/multiprocessing/process.py", line 121, in start
  File "/opt/Apache83/lib/python3.11/multiprocessing/context.py", line 224, in _Popen
  File "/opt/Apache83/lib/python3.11/multiprocessing/context.py", line 281, in _Popen
  File "/opt/Apache83/lib/python3.11/multiprocessing/popen_fork.py", line 19, in __init__
  File "/opt/Apache83/lib/python3.11/multiprocessing/popen_fork.py", line 65, in _launch
OSError: [Errno 24] Too many open files

</code>

mvanhorn:
## Summary

Downloading a very long playlist eventually crashes MeTube with `OSError: [Errno 24] Too many open files`, and the process self-terminates. This releases the per-download `status_queue` manager proxy when a download finishes, so its socket file descriptor is torn down instead of leaking one per completed item.

## Background

andbir reported this in #485, and two commenters independently reproduced the FD exhaustion — one on kernel 6.6.44, which rules out the "old Linux kernel" theory and points at a real leak in MeTube. Each `Download` creates its own `multiprocessing.Manager().Queue()` proxy (`self.status_queue`), and every completed download stays live in the in-memory `PersistentQueue.dict` (`self.done`). Those retained objects keep their Queue proxy, whose client connection to the manager holds a socket FD open. `Download.close()` closed only `self.proc` and never dropped `self.status_queue`, so FDs accumulate across a long playlist until the limit is hit.

`close()` now clears `self.status_queue` in a `finally` block (so the reference is dropped even if `self.proc.close()` raises, and when no process was ever started), letting the proxy be garbage-collected — `Connection.__del__` then closes the FD. This is safe because `close()` runs only after `start()` has awaited `self.status_task`, so no reader remains, and a re-started download reassigns a fresh queue. Added tests for the normal, no-process, and failing-`proc.close()` paths; `pytest app/tests/test_download_queue.py` passes (34 tests).

Closes #485

### DIFF
--- a/app/tests/test_download_queue.py
+++ b/app/tests/test_download_queue.py
@@ -898,6 +898,40 @@ def _make_download(dq_env, *, download_type="video", status="downloading", filen
     )
 
 
+def test_download_close_releases_status_queue(dq_env):
+    download = _make_download(dq_env)
+    status_queue = MagicMock()
+    proc = MagicMock()
+    download.status_queue = status_queue
+    download.proc = proc
+
+    download.close()
+
+    proc.close.assert_called_once()
+    assert download.status_queue is None
+
+
+def test_download_close_releases_status_queue_without_process(dq_env):
+    download = _make_download(dq_env)
+    download.status_queue = MagicMock()
+
+    download.close()
+
+    assert download.status_queue is None
+
+
+def test_download_close_releases_status_queue_when_process_close_fails(dq_env):
+    download = _make_download(dq_env)
+    download.status_queue = MagicMock()
+    download.proc = MagicMock()
+    download.proc.close.side_effect = RuntimeError('close failed')
+
+    with pytest.raises(RuntimeError, match='close failed'):
+        download.close()
+
+    assert download.status_queue is None
+
+
 @pytest.mark.asyncio
 async def test_post_download_cleanup_clears_filename_on_error(dq_env):
     notifier = AsyncMock()
--- a/app/ytdl.py
+++ b/app/ytdl.py
@@ -758,8 +758,11 @@ class Download:
 
     def close(self):
         log.info(f"Closing download process for: {self.info.title}")
-        if self.started():
-            self.proc.close()
+        try:
+            if self.started():
+                self.proc.close()
+        finally:
+            self.status_queue = None
 
     def running(self):
         try:


alexta69:
Merged — thank you, this is a textbook fix. The diagnosis is exactly right: `manager.Queue()` hands back a proxy that holds a live connection to the Manager process, and because a finished `Download` stays in the done list, every completed download was permanently pinning one such connection — so the descriptor count climbed until the instance hit its limit and self-terminated (#485). Nulling `status_queue` in `close()` drops the last reference so the proxy is collected and its connection torn down; `close()` already runs in `_post_download_cleanup` right before the object goes into `done`, so the timing is correct and nothing reads the queue afterward. The `try/finally` is the right touch. I appreciate the three focused tests across the with-process / without-process / close-fails paths. Closed #485 with this — I'll ask the #980 reporter to retest once a release with it ships.

mvanhorn:
Grateful for the merge here. Long playlists should stop hitting the open-file limit now, thanks for looking it over.

PR #1038, opened 2026-07-20
feat: Added handling for unsupported URL

streamer1122:
Unsupported URLs sent from browser plugin or mobile fail with an error only presenting in the log. 
Change committed to present the entry in the UI

### DIFF
--- a/app/tests/test_download_queue.py
+++ b/app/tests/test_download_queue.py
@@ -116,6 +116,54 @@ async def test_add_single_video_goes_to_pending_when_auto_start_false(dq_env):
 
 
 @pytest.mark.asyncio
+async def test_add_unsupported_url_recorded_as_failed_entry(dq_env):
+    """An unsupported/unextractable URL must show up as a red-cross entry in the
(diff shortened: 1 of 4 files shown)


alexta69:
Merged — thank you, this closes a real gap: an unsupported / SSRF-rejected / unextractable URL now lands in Completed as a proper error entry with retry and the detail panel, instead of vanishing after a toast. The `__record_add_failure` wiring on both failure branches is clean, keying by `info.url` is right, and I'm glad you covered the SSRF-reject path too — `_short_title_for_failed_url` with the hostname/raw fallback and the malformed-IPv6 guard is a nice touch.

One adjustment I made while merging: I kept the backend and tests but dropped the `ui/src/app/app.html` hunk that removed the "Click for details" hint. That affordance was added deliberately (#143) to make error rows discoverable as expandable, and I'd rather not remove it for every error entry as a side effect of this change. If you think it reads as noise now that more rows can be errors, I'm happy to consider it — but as its own small UI change we can weigh on the merits. Everything else is in as-is. Thanks!

PR #1039, opened 2026-07-21
Add reusable NAS deployment profile

hunky89:
## What changed

- add a parameterized Docker Compose profile under `deploy/nas`
- keep NAS paths, UID/GID, ports, proxy endpoints, and yt-dlp options in an untracked `.env` file
- route Bilibili domains directly while allowing YouTube and other sites to use an optional LAN proxy
- enable configurable yt-dlp nightly updates for extractor compatibility
- document setup, validation, troubleshooting, updates, rollback, cookies, and network exposure considerations
- link the NAS guide from the main README

## Why

A working NAS deployment previously depended on machine-specific Compose files with hard-coded LAN addresses and storage paths. When the proxy host address changed, both YouTube and Bilibili downloads failed. Bilibili also required a newer yt-dlp nightly after its WBI API behavior changed.

This profile captures the validated deployment pattern without committing personal infrastructure details, so it can be reused on another Linux NAS by copying `.env.example` to `.env` and setting environment-specific values.

## Impact

No backend or frontend behavior changes. Existing deployments are unaffected. NAS users gain an optional, documented deployment profile.

## Validation

- `docker compose config --quiet` passed on a Linux NAS running Docker Compose v5.1.4
- rendered Compose configuration was inspected for correct JSON and environment interpolation
- YouTube and Bilibili extraction and real downloads were verified on the source NAS
- `git diff --check` passed
- README remains below the documented 25,000-character limit
- scanned the committed files for personal IPs, paths, usernames, email addresses, and tokens


alexta69:
Thank you for putting this together, and sorry it sat as a draft for a while without a clear answer from me.

I am going to decline it, on scope rather than quality. A deployment profile is configuration, not application code, and once it lives in this repo it becomes something I have to keep correct: the NAS paths, the UID/GID conventions, the Bilibili direct-route rules and the proxy split are all specific to one setup, and every future change to MeTube would have to be checked against a compose file I cannot test on hardware I do not have. That is a maintenance commitment I am not able to honour, and a stale deployment profile in the official repo is worse than none.

The repo already carries a minimal `docker-compose.yml` as the starting point, and the parts of your setup that generalise are a better fit for the [wiki](https://github.com/alexta69/metube/wiki), where recipes can be attributed to their author and updated without a release. If you would like to write it up there — particularly the routing split and the nightly-update configuration, which are the genuinely non-obvious bits — I would be glad to link it, the same way the VAAPI recipe from #1045 went in.

None of this reflects on the work itself; it is a question of what I can commit to maintaining.

PR #1041, opened 2026-07-24
Fix playlist folder preservation when retrying failed downloads

jahruz67:
## Summary

- Add a server-side retry endpoint for failed downloads.
- Preserve playlist and channel metadata when retrying individual items.
- Ensure retried playlist videos and music download into their original playlist folder instead of the root downloads directory.
- Persist only the compact metadata needed for retries across server restarts.
- Keep successful completed records free of extractor metadata.

## Testing

- Added backend, API, persistence, and frontend regression tests.
- 86 targeted backend tests pass.

**PR update:**

**Fix: Preserve `track_number` across download retries**

When a download from a playlist/album fails and the user retries it, the `track_number` was being lost because `_compact_persisted_entry()` in `ytdl.py` only preserved keys starting with `playlist` or `channel`, plus a small whitelist. The `track_number` field didn't match any of these prefixes, so it was dropped during persistence. On retry, `MusicMetadataPreProcessor` couldn't detect an album signal and the retried download had no track number tag.

Fixed by adding `"track_number"` to the `_COMPACT_ENTRY_EXTRA_KEYS` frozenset. This is a single-line change that keeps the persisted entry compact while ensuring the track number survives the failure/retry cycle.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -893,6 +893,16 @@ async def cancel_add(request):
     return web.Response(text=serializer.encode({'status': 'ok'}), content_type='application/json')
 
 
+@routes.post(config.URL_PREFIX + 'retry')
+async def retry(request):
+    post = await _read_json_request(request)
+    ids = _require_id_list(post)
+    if len(ids) != 1:
+        raise web.HTTPBadRequest(reason="'ids' must contain exactly one download id")
+    status = await dqueue.retry(ids[0])
+    return web.Response(text=serializer.encode(status), content_type='application/json')
+
+
 @routes.post(config.URL_PREFIX + 'subscribe')
 async def subscribe(request):
     post = await _read_json_request(request)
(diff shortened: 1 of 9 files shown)


alexta69:
Merged, thanks — and good root-cause work. Since the public download dict deliberately excludes `entry`, the client-side retry could never have carried playlist context across a retry, so moving it server-side was the right call.

I applied two fixes at merge:

1. `retry_entry` was dropped when `__add_entry` recursed on a `_type: url` / `url_transparent` result. Extraction runs with `extract_flat=True`, so that path is live, and a retried playlist item taking it fell back to `OUTPUT_TEMPLATE` and landed in the root directory — the same bug this PR fixes, just on the other branch. It's carried through now, with a regression test.

2. `retry()` called `dqueue.add()` directly, which bypasses the `parse_download_options` gates that `/add` applies. Stored `ytdl_options_overrides` were re-applied even after `ALLOW_YTDL_OPTIONS_OVERRIDES` had been turned off, and preset names since removed from the configuration still passed through. Both are re-checked against the current configuration at retry time now.

I also changed the endpoint to take `{"id": "..."}` rather than `{"ids": ["..."]}`. It rejected anything but exactly one id, so the plural form advertised a batch that didn't exist. Unlike `/delete` and `/start` — which act on local state and can't meaningfully fail for one id and not another — a retry re-extracts the URL and can fail per item, and the caller removes that item's done record only once it's confirmed queued. A real batch form would need per-id results in the response, which is only worth designing alongside moving done-record deletion server-side.

The `track_number` catch was a good one too — that would have quietly undone the #1031 metadata on any failure/retry cycle.

PR #1056, opened 2026-08-16
fix: surface yt-dlp warning context on failed downloads

### LINKED ISSUE #1047: [Feature]: Clearer error message
### Prerequisites

- [x] I have searched existing issues and checked the "Already decided" list above
- [x] I have considered if this feature belongs in yt-dlp instead (downloading, processing, site support) — if so, please request it in the [yt-dlp repository](https://github.com/yt-dlp/yt-dlp/issues)

### Feature Description

Hi!

I would love to see a bit more valuable error messages. Example:

I tried downloading this video:
https://www.youtube.com/watch?v=u2HSc2Ym1Vk

In MeTube, I received the error message

```
Title: Andreas Bourani - Voll gerne (aus "Vaiana")
URL: https://www.youtube.com/watch?v=u2HSc2Ym1Vk
Message: ERROR: [youtube] u2HSc2Ym1Vk: No video formats found!
```

I googled the error and found https://github.com/alexta69/metube/issues/881 and https://github.com/alexta69/metube/issues/537

Both led me in the wrong direction, though.
I then opened MeTube logs and found this messsage:

```
metube  | INFO:ytdl:Starting download for: Andreas Bourani - Voll gerne (aus "Vaiana") (https://www.youtube.com/watch?v=u2HSc2Ym1Vk)
metube  | WARNING: [youtube] Video unavailable. This video contains content from bryhuangpub, who has blocked it from display on this website or application Watch on YouTube
metube  | WARNING: Only images are available for download. use --list-formats to see them
metube  | WARNING: Requested format is not available
metube  | ERROR: [youtube] u2HSc2Ym1Vk: No video formats found!
```

So the error is neither MeTube nor yt-dlp, but on YouTube config side.

I checked out another video and it worked fine, eg.
https://www.youtube.com/watch?v=5RrXNp-7KrM

So my feature request is: can you please add "better" error messages, maybe in a pop-out or something? :-)

### Implementation Plan

No, I'm requesting this feature but cannot implement it

### Additional Context

_No response_

mvanhorn:
When yt-dlp explains a download failure through warnings before returning `No video formats found`, MeTube currently discards that explanatory context and exposes only the final generic error. The existing `Download._download` path also emits an error status without a `msg` whenever yt-dlp returns a nonzero result instead of raising `YoutubeDLError`, leaving the completed entry with no explanation at all. The maintainer accepted a backend-only fix that reuses yt-dlp's existing logger output, requires no dependency or UI changes, and preserves the current failed-entry detail flow through `DownloadInfo.msg`.

## Summary

Add a download-scoped yt-dlp logger in `app/ytdl.py` that forwards debug, warning, and error records to MeTube's existing `ytdl` logger while retaining non-empty warning messages in their emitted order for failure reporting. Pass that logger through the existing `ytdl_params` in `Download._download`, and when `YoutubeDL.download()` returns nonzero, enqueue an error status whose `msg` contains the captured warning context; if no warning was emitted, provide a concise non-empty fallback tied to the failed download result. Keep the current `YoutubeDLError` exception message path and successful-download status unchanged, and do not add platform-specific parsing or a new frontend surface.

## Testing

- A fake yt-dlp download emits the uploader-blocked and format warnings from the report, then returns nonzero; the queued terminal status is `error`, its message preserves the useful warning context, and those warnings are still forwarded to server logging.
- yt-dlp returns nonzero without emitting a warning; the queued terminal error has a stable, non-empty fallback message rather than a missing `msg`.
- yt-dlp emits a warning but returns zero; the terminal status remains `finished` and warning text is not attached as a failure message.
- yt-dlp raises `YoutubeDLError`; MeTube continues to enqueue the exception text as the error message without replacing it with the nonzero-return fallback.
- Logger debug/error methods preserve existing level routing and tolerate yt-dlp's message calls without changing download status on their own.

Fixes #1047

### DIFF
--- a/app/tests/test_ytdl_utils.py
+++ b/app/tests/test_ytdl_utils.py
@@ -450,6 +450,65 @@ class DownloadLoggerTests(unittest.TestCase):
         self.assertIn('WARNING:ytdl: useful warning ', logs.output)
         self.assertIn('ERROR:ytdl:error detail', logs.output)
 
+    def test_retains_only_the_last_distinct_warnings(self):
+        logger = ytdl._DownloadYtdlLogger()
+        cap = ytdl._MAX_RETAINED_WARNINGS
+
+        with self.assertLogs('ytdl', level='WARNING') as logs:
+            for index in range(cap + 3):
+                logger.warning(f'fragment {index} not found')
+
+        self.assertEqual(
+            logger.warnings,
+            [f'fragment {index} not found' for index in range(3, cap + 3)],
+        )
+        # Every warning still reaches the log; only the retained list is bounded.
+        self.assertEqual(len(logs.output), cap + 3)
+
+    def test_repeated_warning_is_retained_once(self):
+        logger = ytdl._DownloadYtdlLogger()
+
+        with self.assertLogs('ytdl', level='WARNING'):
+            logger.warning('Requested format is not available')
+            logger.warning('Only images are available for download')
+            logger.warning('Requested format is not available')
+
+        self.assertEqual(
+            logger.warnings,
+            ['Requested format is not available', 'Only images are available for download'],
+        )
+
+    def test_failure_message_puts_the_error_last(self):
+        logger = ytdl._DownloadYtdlLogger()
+
+        with self.assertLogs('ytdl', level='WARNING'):
+            logger.warning('Only images are available for download')
+
+        self.assertEqual(
+            logger.failure_message('ERROR: [youtube] u2HSc2Ym1Vk: No video formats found!'),
+            'Only images are available for download\n'
+            'ERROR: [youtube] u2HSc2Ym1Vk: No video formats found!',
+        )
+
+    def test_failure_message_skips_a_last_warning_that_repeats_the_error(self):
+        logger = ytdl._DownloadYtdlLogger()
+
+        with self.assertLogs('ytdl', level='WARNING'):
+            logger.warning('Video unavailable')
+            # yt-dlp labels errors but hands warnings to the logger unlabelled,
+            # so the same text can arrive through both routes.
+            logger.warning('Requested format is not available')
+
+        self.assertEqual(
+            logger.failure_message('ERROR: Requested format is not available'),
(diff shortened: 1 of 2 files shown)


alexta69:
Thanks for picking this up so quickly, and for the tests — the logger design is right, the routing is clean, and the fallback for a nonzero return with no warnings is a genuine improvement on its own.

There is one problem though, and it is the important one: **as written, this does not fix the case #1047 reports.**

MeTube never sets `ignoreerrors`, so when yt-dlp hits `No video formats found!` its `trouble()` path raises `DownloadError` rather than returning nonzero. Momro's failure came out through the `except yt_dlp.utils.YoutubeDLError` branch — the `ERROR: [youtube] u2HSc2Ym1Vk: No video formats found!` text he quoted is exactly `str(exc)`. Your change attaches the captured warnings only on the `ret != 0` branch, and the exception branch still reports `str(exc)` alone, so the explanatory warnings are dropped in precisely the scenario that prompted the issue. `test_youtube_dl_error_message_takes_precedence_over_warnings` currently locks that behaviour in.

What I think it needs: the exception path should also carry the warning context, with the exception text kept as the final line so the actual error stays prominent. Worth checking whether the last warning duplicates the error text and skipping it if so.

Second thing, independent of the above: **`warnings` is unbounded.** Fragmented and live downloads can emit a warning per fragment, and that list is joined into `msg`, persisted in the completed queue and broadcast to every connected client. A long recording could produce a very large message. Please cap it — keeping the last N distinct warnings would do.

Minor: `ytdl_logger` is bound inside the `try`, so if it is referenced from the `except` block it needs to be initialised before it.

Happy to merge once the exception path carries the context and the list is bounded.

mvanhorn:
@alexta69 the exception path now carries the warning context.

`DownloadError` / `YoutubeDLError` uses the same bounded list as the nonzero-return branch: last 5 distinct warnings, then `str(exc)` as the final line. If the last warning is the error text with the ERROR:/WARNING: label stripped, it's dropped so it doesn't repeat. Logger is created before the try.

The old test that locked in exception-text-only is gone.

Issue #133, opened 2022-03-31
Android app


binarygeek119:
Would it be possible to make a android app for submitting videos

PikuZheng:
I'm using ifttt... but can you open metube webpage with your phone's web browser?

alexta69:
I suppose some sort of solution can be arrived at with IFTTT or a similar tool (I guess you want to click "Share" in the YouTube app and send the video to MeTube?)
Leaving the issue open in case someone wants to share their config of how they're doing it. I don't think an Android app is coming though.

bdr99:
The Android app Termux has a feature allowing you to create a shell script to act as a share target for other apps. The shell script will receive the shared URL as the first parameter. I installed Termux and created a script at `bin/termux-url-opener` with the following contents:
```sh
curl -X POST https://my-metube-server.my-domain.com/add -H 'Content-Type: application/json' -d "{\"url\":\"$1\",\"quality\":\"1440p\"}"
```

Now, I can open the YouTube app and share any video with Termux, and MeTube will download it 😀

cochraneserver:
I create a specific YouTube playlist I want either video or audio downloaded from, go to MeTube and post the Unlisted Playlist link

husudosu:
I'm interested on building an Android apps. I see you use Socket.IO and few HTTP requests. I can build the app by using Ionic Vue under MIT license, if it's okay with you @alexta69

alexta69:
@husudosu of course it's perfectly OK with me! It'll be a terrific addition. Thanks! 😌

cochraneserver:
I'd use an android app if it was available:)

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

husudosu:
I've created this app to manage Metube from Android. It supports sharing video links from other apps like Youtube. Currently it's on alpha stage, most of the error handling missing need to do optimizations etc...

- [Repo](https://github.com/husudosu/metube-app)
- [First build available here](https://github.com/husudosu/metube-app/releases/tag/0.0.1)

When you configure Metube server address be sure to add HTTP/HTTPS and include your port too!

[I made a video demonstration of current state.](https://www.youtube.com/watch?v=YrNinKEP8jk)

1RandomDev:
I guess I'm a bit late.
But anyway I've also created a simple app that lets you add new downloads and share videos from YouTube/other apps with MeTube.
[Here](https://github.com/1RandomDev/metube-android) you can find the code and the first release.

[...]

PR #1077, opened 2026-09-13
docs: add MeTube Mobile (Android) to Sending links to MeTube

sagheerys:
Adds one line under "Sending links to MeTube" for MeTube Mobile, two open-source (GPL-3.0) Android apps that send links from the share menu. Lite pulls each finished download to the phone and then deletes it from the server; Super keeps files on the server and streams them.

Signed APKs are on the releases page, and the project states it is unofficial. The code was written with heavy AI assistance, and most commits carry a Claude co-author line.

The README stays under the 25,000 character limit.

If you'd prefer different wording or placement, or want the older Android apps from #133 listed as well, I'm happy to change it. Thanks for MeTube, my family uses it every day.


alexta69:
Merged, thanks — that section is where people look for exactly this.

One thing for anyone finding this later: the Lite app removes each download from the server once it has pulled it to the phone, which is the behaviour you want or very much isn't, depending on why you run MeTube. Your README wording says so and I've kept it as you wrote it.

I'll leave the #133 apps out; one current entry per platform is what that section can carry.

PR #1078, opened 2026-09-14
feat: add direct streaming routes /watch and /dl/<name>.<ext>

fgilde:
## What

Two synchronous GET routes for the "just give me this one thing right now" case. They hand a URL, a bare video ID or a search term to yt-dlp, download that single item into a per-request temporary directory, stream the file straight back and delete the directory once the body has been sent. Nothing is written to `DOWNLOAD_DIR`, nothing enters the queue or the history, nothing is cached.

| Route | Result |
|-------|--------|
| `GET /watch?v=<id-or-url>[&download=1]` | The video as mp4 (up to 1080p). Mirrors YouTube's own URL shape, so a link only needs its host swapped. `v` is passed to yt-dlp verbatim, which accepts bare IDs itself; nothing YouTube-specific is parsed. |
| `GET /dl/<name>.<mp4\|mp3\|jpg>[?url=…][&download=1][&ts=90]` | The extension picks the format. Without `url`, `<name>` becomes a `ytsearch1:` query (`-`, `_`, `+`, `.` turn into spaces); with `url`, that URL is fetched and playlists collapse to their first item. |

`download=1` sends `Content-Disposition: attachment`, otherwise `inline` so the browser plays it. Range requests work (`206`), so `<video>` seeking is fine. Titles go into the file name with an ASCII fallback plus RFC 5987 `filename*`.

`jpg` without `ts` returns the extractor's thumbnail (the same `skip_download`/`writethumbnail` path the form uses). With `ts` (`90`, `1:30`, `1m30s`), yt-dlp resolves a video-only mp4 stream without downloading it and ffmpeg seeks to `ts` on that URL and decodes one frame.

```
/watch?v=PVtSAXSv7Fo
/watch?v=PVtSAXSv7Fo&download=1
/dl/Greenday-Basketcase.mp3
/dl/something.mp4?url=https://youtu.be/PVtSAXSv7Fo
/dl/whatever.jpg?url=https://youtu.be/PVtSAXSv7Fo&ts=42
```

## How

- `app/direct.py` (new): name/`ts` parsing, `Content-Disposition` builder, the `fetch()` worker, and `TempFileResponse`, a `FileResponse` that removes its temp dir after sending.
- `app/main.py`: the two routes, `_direct_serve` (validate, semaphore, per-request process, error mapping to `400`/`502`), `robots.txt` now disallows `/dl/` and `/watch`.
- Format selection reuses `dl_formats.get_format` / `get_opts`, so an mp3 comes out exactly like an audio download from the form (extracted audio, embedded thumbnail, metadata). mp4 uses the form's mp4 selector at 1080p with a `/bv*+ba/b` fallback and `merge_output_format=mp4`.
- `YTDL_OPTIONS` including the runtime `cookiefile` override and `impersonate` is layered underneath; presets and per-download overrides are not exposed.
- Concurrency is capped by a semaphore sized by `MAX_CONCURRENT_DOWNLOADS`; extra requests wait.

## Security

- `validate_url` runs on `v`/`url` before anything is spawned; `install_socket_guard` runs inside the worker process, so redirects and manifest-derived media URLs are re-checked at connect time. Same two layers as a queue download; `ALLOW_PRIVATE_ADDRESSES` disables both as elsewhere.
- Each fetch runs in a throwaway single-worker `ProcessPoolExecutor` so the guard can be installed without touching the server process.
- Output paths go through `_ConfinedYoutubeDL` with the temp dir as the only allowed root; the served file is always `media.<ext>` inside it. `<name>` never touches the filesystem.
- The ffmpeg frame grab fetches the media URL outside Python's socket module, so only `validate_url` covers that single fetch. This is the same gap the existing clip feature has (yt-dlp's `FFmpegFD` fetches the same way).

## Scope

No new environment variables. The 1080p cap, the 720p frame source and the reuse of `MAX_CONCURRENT_DOWNLOADS` are fixed defaults. If a smaller first step is preferred, `/watch` plus `/dl/*.mp4|mp3` is the core and the `jpg`/`ts` frame grab can be split into a follow-up.

## Verification

- `python -m compileall app`, `uv run pytest app/tests/` (10 new tests in `app/tests/test_direct.py`; the worker is mocked out there).
- End-to-end against YouTube on a running instance: mp4 via `/watch` including a `206` range request, mp3 via search as attachment, thumbnail, and the frame at `ts=42` checked visually. Error paths return `400` (bad extension, missing `v`, refused internal URL) and `502` with `yt-dlp: <message>` (unavailable video). No `metube-direct-*` directory was left behind.
- README stays under the Docker Hub limit (24,878 chars).


alexta69:
I'm going to decline this, and it's about what MeTube would be signing up to own rather than about the state of the code.

These routes are a second product sharing a process with the first. Nothing enters the queue, the history, or `DOWNLOAD_DIR`; the file is streamed and the directory deleted. Every surface MeTube has — the queue, folders, subscriptions, the file server, the API — is bypassed rather than used.

The concrete problems that follow from that:

**MeTube ships with no authentication**, on the assumption that an operator who exposes it puts a reverse proxy in front. Today an anonymous request leaves a queue row and a file the operator can see. `/watch?v=` is a fetch that streams bytes to the caller and leaves nothing behind — same reachability, no record. That's a different exposure for every instance that's already on the internet, and it arrives switched on, with no flag to opt into.

**`/dl/Greenday-Basketcase.mp3` turning into `ytsearch1:` is inference on input the user can't inspect.** I declined #1068 on the same grounds, and it's worse here: at least a queue row shows you what you got. A URL shape that silently returns *a* file matching a guess at your filename is one a user can't verify. No tool in this space does this — yt-dlp makes you ask for `--default-search`, and JDownloader stages ambiguous input for confirmation.

**The surface is permanent.** RFC 5987 filename encoding, range requests, a timestamp mini-language with three accepted forms, an ffmpeg single-frame grab, a per-request process pool, and a second format-selection path that has to track `dl_formats` as it changes. You've already documented one security gap in it — the frame grab fetching outside Python's socket module, where the connect-time guard can't see it. That's the maintenance shape of the whole thing.

For your own use, the existing API already does the durable half: `POST /add`, then fetch the finished file from `/download/<path>`. And for the transient half — stream it, keep nothing — `yt-dlp -o -` in a small handler in front of MeTube is a closer fit than building it inside, because it doesn't have to answer to anything else in the process.

